Issue #610, opened 2014-09-08
isoline don't have labels


potter2:
Isoline plots don't have labels when requested.

doutriaux1:
fixed by @dlonie already in

PR #926, opened 2014-11-25
Enable labels for isolines 

aashish24:
#610


dlonie:
@doutriaux1 Can you point me at the where text properties for the labels are stored in vcs? I'm looking into hooking them into the mapper now.

doutriaux1:
@dlonie I think we need to update isofill. Here is how it works for isolines. I think we will need to add the "text" bit to isofill
```
In [3]: iso=x.createisoline()

In [4]: iso.list()
 ----------Isoline (Gi) member (attribute) listings ----------
graphics method = Gi
name = __isoline_795667460274169
projection = linear
xticlabels1 = *
xticlabels2 = *
xmtics1 = 
xmtics2 = 
yticlabels1 = *
yticlabels2 = *
ymtics1 =  
ymtics2 =  
datawc_x1 = 1e+20
datawc_y1 =  1e+20
datawc_x2 =  1e+20
datawc_y2 =  1e+20
datawc_timeunits =  days since 2000
datawc_calendar =  135441
xaxisconvert =  linear
yaxisconvert =  linear
label =  n
line =  ['solid']
linecolors =  [1]
linewidths =  [1.0]
text =  None
textcolors =  None
level =  [[0.0, 1.0000000200408773e+20]]
clockwise =  [0]
scale =  [1.0]
angle =  [35.0]
spacing =  [1.0]
```
Looking at older version. Only textcolor is respected and it applies the first color to all labels.

So what we really need is two attributes (we can keep textcolor for backward compatibility but feel free to yank it I don't think anybody uses it)

texttables=[...] # for each level use the texttable objects list here (we can default to None which would use "default" texttable
textorientations=[..] same ting for orientation.

If you look into the plot calls for a text object you can see how I map texttables and textorientations to VTK text objects.

dlonie:
So we're adding new functionality, then? I thought we were focusing on reproducing the existing functionality at this point.

I'm not sure those new members are needed -- from what I can see, the text table objects define a font size, face, and possibly a color (from looking at prepTextProperty). Would each level really require a different font size / face? I think we'd want these consistent for all levels. Same with orientations -- these labels should always be centered on the anchor point and aligned with the isoline.

Perhaps a single text table and array of colors would be sufficient? (I think that's what is already there -- not sure what the `None` represents for `text`, but I assume a texttable?)

doutriaux1:
I agree we could make the textable/orientation not lists they all should be the same, with the exception of colors. Since each isoline can be of a different color it probably makes sense to allow each label to be of the matching color.

doutriaux1:
Ok let's not add these right away if you want, you can split it in two, but I don't think it's much work to add these two attributes and it will be easier to do it now rather than later.

doutriaux1:
do you want me to take a stab at adding the attributes on the isoline method? And you just do the VTK implementation?

dlonie:
I'm confused, your first reply seems to agree that these aren't needed, and the other two sound like they are. Do we need them?

It's not a ton of work to add these attributes, I'm just not sure it makes sense, particularly for orientations as all labels should always be centered horizontally and vertically, and rotated to follow the isoline.

If there's a need for differing the font size / face for each level of the isoline, we can certainly do that, but it seems odd that this would be a desired feature.

doutriaux1:
what's really need it font type/size and color. I agree angle and such will probably never be used (although I can see the use of the 90degree) center/half will probably not be used either, but they're there as part of the textable and textorientation so we might as well respect them to make the experience consistent accross vcs. 
I think we should have them ultimately, and probably now since it's not much work to do it now vs later.
But I will leave it up to you to decide if you want to do it all at once or split it and open an issue for the second part.

dlonie:
Ok, I'll modify the mapper to allow multiple text properties for level. This way the face, size, and color can be set independently for each level.

I can take care of the VTK side of things, but I'd rather you handle the VCS changes (if you don't mind), as you're more familiar with what needs to be updated and where.

dlonie:
vtkLabeledContourMapper now supports per-isoline text properties in our VTK fork:
UV-CDAT/VTK@f26f2954fc746c0914271241a43e7036dc3978cb

When the VCS structures are in place, I can hook them up in VTKPlots.py.

doutriaux1:
ok will put the vcs plugs in place early next week. Thanks!

aashish24:
I am going to push a branch that will update the VTK. What else we need here?

dlonie:
VTK is ready to go. Once vcs has implemented the structures to initialize the new VTK mapper, I'll hook them up and we should be done with this issue.

doutriaux1:
last I remember it was not working, please please please don't merge untiI chekced it works properly.
@dlonie let me know when VTK has been updated I'll rerun this branch and see if it's ok. Thanks.

dlonie:
The VTK functionality is already merged into the UVCDAT fork -- UV-CDAT/VTK@f26f295. All we need are the per-contour text properties in vcs so we can connect them to the VTK layer.

pletzer:
Please unsubscribe pletzer@txcorp.com.  Alex do not work at Tech-X anymore.
Thanks,
Sveta

doutriaux1:
@aashish24 @dlonie 
Requires https://github.com/UV-CDAT/VTK/tree/contour-labels
but that conflicts with uvcdat-master can you please take a look? thx. Using contour labels for now in the mean time.

dlonie:
Just use contour-labels for now. When we bump VTK, all will be resolved :)

On Thu, Jan 29, 2015 at 2:29 PM, Charles Doutriaux <notifications@github.com

doutriaux1:
Yes that's what I said. BTW when are we finally bumping VTK?

doutriaux1:
well... something's still off
```
  File "test_clt_isolines.py", line 12, in <module>
    x.plot(s,gm)
  File "/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/Canvas.py", line 2593, in plot
    a = self.__plot( arglist, keyargs )
  File "/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/Canvas.py", line 3653, in __plot
    returned_kargs = self.backend.plot(*arglist,**keyargs)
  File "/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/VTKPlots.py", line 396, in plot
    returned.update(self.plot2D(data1,data2,tpl,gm,vtk_backend_grid=vtk_backend_grid,vtk_backend_geo=vtk_backend_geo))
  File "/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/VTKPlots.py", line 894, in plot2D
    mapper.GetTextProperty().SetFontSize(12);
AttributeError: GetTextProperty
```
But I did merge master into the branch will go back to the state this branch is at

doutriaux1:
nope still fails for me....

doutriaux1:
here's my basic test that fails:
```
import vcs
import sys,os,cdms2

f=cdms2.open(os.path.join(sys.prefix,"sample_data","clt.nc"))

s=f("clt")

x=vcs.init()

gm=x.createisoline()
gm.labels="y"
x.plot(s,gm)
raw_input("Press enter")
```

dlonie:
Ah, it looks like I didn't update that part of vcs when I added the per-line font properties.

Feel free to just comment out the problematic lines while working on the vcs additions. We'll need to rewrite that bit anyway to create the new text properties, so I'll fix that up in a bit (I'm on a different project at the moment, or I'd fix it now...)

dlonie:
Just checking on this -- are the new text property arrays ready in VCS?

doutriaux1:
thanks for pinging! I need to do this asap!

doutriaux1:
@dlonie ok that's still not in uvcdat-master right? I still still to build the special one, correct?

dlonie:
UVCDAT's VTK master has the new API for multiple text properties, but all of the work that enables it in vcs is in the special branch.

doutriaux1:
@dlonie ok I'm back onto this one...
I merged master back in. It's using latest VTK
Now I'm running:
```
import vcs
import sys,os,cdms2
f=cdms2.open(os.path.join(sys.prefix,"sample_data","clt.nc"))
s=f("clt")
x=vcs.init()
gm=x.createisoline()
gm.labels="y"
x.plot(s,gm)
raw_input("Press enter")
```
Out of the box I get:
```
  File "/home/doutriaux1/test_labels.py", line 12, in <module>
    x.plot(s,gm)
  File "/lgm/uvcdat/2015-02-05/lib/python2.7/site-packages/vcs/Canvas.py", line 2593, in plot
    a = self.__plot( arglist, keyargs )
  File "/lgm/uvcdat/2015-02-05/lib/python2.7/site-packages/vcs/Canvas.py", line 3653, in __plot
    returned_kargs = self.backend.plot(*arglist,**keyargs)
  File "/lgm/uvcdat/2015-02-05/lib/python2.7/site-packages/vcs/VTKPlots.py", line 396, in plot
    returned.update(self.plot2D(data1,data2,tpl,gm,vtk_backend_grid=vtk_backend_grid,vtk_backend_geo=vtk_backend_geo))
  File "/lgm/uvcdat/2015-02-05/lib/python2.7/site-packages/vcs/VTKPlots.py", line 894, in plot2D
    mapper.GetTextProperty().SetFontSize(12);
AttributeError: GetTextProperty
```
So per your comments above I commented these two lines (circa ine894):
```
        #mapper.GetTextProperty().SetFontSize(12);
        #mapper.GetTextProperty().SetColor(0, 0, 0)
```

Now I get a blank plot... 
blank

dlonie:
I'll take a look once the new font properties are in.

doutriaux1:
@dlonie sorry it took so long. I'm looking at the whole thread right now. Labels are only for isolines. So there is no need to add anything to vcs. All the needed components are already in place. 
isoline.text is a list of the text objects to use (the color should be overwritten if isoline.textcolor exists.

in short 
use the "text" i.e. text combined object defined in the isoline.text list, use the properties from these object

tiny example of test code:

```
import vcs
import sys,os,cdms2
f=cdms2.open(os.path.join(sys.prefix,"sample_data","clt.nc"))
s=f("clt")
x=vcs.init()
gm=x.createisoline()
gm.labels="y"
texts=[]
for i in range(20):
  t=x.createtext()
  t.color=230+i
  texts.append(t.name) # i think it can take either text object or text name
gm.text = texts
vcs.gettext(texts[0]).list()
x.plot(s,gm)
```

aashish24:
@dlonie can we merge this one now or do we need some more fixes?

dlonie:
Still need to hook up the text properties. Should be soon.

doutriaux1:
just so you know, this absolutely does not work for me. Even commenting out textproperties

dlonie:
Yes, vcs needs to catch up with vtk changes. I'll have them in today.

dlonie:
When I run the script above and add

```        print "gm.text =",gm.text
        print "gm.textcolors =",gm.textcolors```

to VTKPlots where we set up the text properties, I get:

```gm.text = []
gm.textcolors = None```

Where do these properties go?

**Comment by doutriaux1** [comment]
thanks @dlonie

**Comment by doutriaux1** [comment]
If empty it means "default"

**Comment by dlonie** [comment]
But shouldn't the script set it up to be not empty?

    gm.text = texts

**Comment by dlonie** [comment]
Also, it looks like the color is being set in the `texts` object, and there is also a `textcolors` object on the isoline -- which should be used?

**Comment by doutriaux1** [comment]
@dlonie I will look to see why it's not setup. textcolor should overwrite text if set (most of the time users only want to set the colors, so it's like a convenience function)

**Comment by doutriaux1** [comment]
@dlonie fixed the iso.text issue but be carefull when you receive it
if can be either:
a text object (combination of text table and text orientation) or a just a text table or text orientation,
In case it's one the the last two, simply assume "default" for the properties of the missing one.

**Comment by doutriaux1** [comment]
@dlonie also merged master into it

**Comment by doutriaux1** [comment]
@dlonie updated test script

```
import vcs
import sys,os,cdms2
f=cdms2.open(os.path.join(sys.prefix,"sample_data","clt.nc"))
s=f("clt")
x=vcs.init()
gm=x.createisoline()
gm.labels="y"
texts=[]
colors = []
for i in range(20):
    t=x.createtext()
    t.color=230+i
    colors.append(100+i)
    if i%2 == 0:
        texts.append(t.name) # i think it can take either text object or text name
    else:
        texts.append(t)
gm.text = texts
gm.textcolors = colors
vcs.gettext(texts[0]).list()
gm.list()
x.plot(s,gm)
```

**Comment by dlonie** [comment]
What would be the best way to use this with the existing prepTextProperty method? Or should a new method be written?

**Comment by doutriaux1** [comment]
@dlonie I'm not quite sure, yes that would be the way. get your property and pass to that along with the texttable and textorientation names. a textcombined simply represent a "union" of two such object
e.g:
```
In [1]: t=x.createtextcombined()

In [2]: t.name
Out[2]: '__texttable_33146519083624:::__textorientation_996084430026894'
```
simply split the name at ":::" and you get both your texttable and textorientation object names

That should work. 

In case where you only get a texttable or textorientation use "default" for the name of the other

**Comment by dlonie** [comment]
Is there a function in vcs that produces some standard object from these many possible text representations? Or is new parsing/conversion code supposed to be written each time a text object is used?

Since we're not really sure what all of the possible forms these text specifications can take, I'm rather uncomfortable writing the code to handle them -- unless there is a standard converter/interpreter for these, it's just a bug waiting to happen.

**Comment by dlonie** [comment]
I've written it to expect a list of textcombined objects, as that's what this sample script produces. If other isoline.text formats are going to be used, we'll need to update this to use a standard method of producing a common representation.

From the sample script:
test

Made the fonts a bit smaller:
test

Without specifying fonts (e.g. using the defaults):
test

**Comment by dlonie** [comment]
Added test and baseline.

Baseline PR: UV-CDAT/uvcdat-testdata#21

**Comment by dlonie** [comment]
This should be ready to merge once everyone is happy with it.

**Comment by doutriaux1** [comment]
that looks great! Let me test it. I think textcombined is good enough. I will tweak the VCS_validation to make sure nothing else is accepted.

**Comment by dlonie** [comment]
Ok -- feel free to strip out the giant `HACK:` warning I added toVTKPlots.py once it's irrelevant ;)

**Comment by doutriaux1** [comment]
@dlonie ok we're getting close
at the moment the labels are always on even if turned off
looking to see how to disable this

**Comment by doutriaux1** [comment]
iso.textcolors is not overwriting
```
tt=x.createtext()
tt.color=243
tt.height=25
iso.text=[tt,]
iso.textcolors=[242,244]
```

**Comment by doutriaux1** [comment]
@dlonie ok I pushed a few changes
1- labels="n" is now respected
2- textcolors actually overwrite
3- you can now pass textcombined texttable or textorientation

File: Packages/vcs/Lib/VTKPlots.py [hunk 1]
```diff
@@ -884,8 +885,64 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
...
+         if gm.text or gm.textcolors:
+          colorOverrides = gm.textcolors if gm.textcolors else [None] * len(gm.text)
+          texts = gm.text if gm.text else [None] * len(gm.textcolors)
+          while len(texts)<Nlevs:
+              texts.append(texts[-1])
+          while len(colorOverrides)<Nlevs:
+              colorOverrides.append(colorOverrides[-1])
+          tprops = vtk.vtkTextPropertyCollection()
+          for tc,colorOverride in zip(texts, colorOverrides):
+              if vcs.queries.istextcombined(tc):
```
dlonie: I still think this should be a core vcs function somewhere -- reproducing logic like these every time it's needed is burdensome, error-prone, and difficult to maintain if changes are made to the text specification (every instance of this logic would need to be updated manually).

Not worth delaying this patch, but something to think about.

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -884,8 +885,64 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
...
+              elif vcs.queries.istexttable(tc):
+                  tt=tc.name
+                  to="default"
+              elif vcs.queries.istextorientation(tc):
+                  to=tc.name
+                  tt="default"
+              if colorOverride is not None:
+                  tt=vcs.createtexttable(None,tt)
+                  tt.color = colorOverride
+                  tt=tt.name
```
dlonie: This doesn't look right to me -- wouldn't this replace the tt object we just configured with a string?

dlonie: nvm, I see now -- the internal tables store it.

We could just pass the override color index to the prepTextProperty method without having to touch the canvas state. I think this would be cleaner, but either way works.

aashish24:
yay!

PR #949, opened 2015-01-06
Issue948 durack1 cdscan update

### LINKED ISSUE #948: cdscan tweaks to work with non CF-compliant input files
Thanks to @painter1, there have been a bunch of tweaks made to cdscan to make it work with quirky CAM/ACME files, these changes deal with data that is not strictly CF-compliant - as detailed by Jeff below:
```
-----Original Message-----
From: "Jeffrey F. Painter"
Date: Wednesday, November 26, 2014 at 3:06 PM
To: "Paul J. Durack"
Cc: "Doutriaux, Charles"
Subject: Re: ACME time axis workaround for cdscan?

I have some b1850c5_t2 climatologies on my Mac, though not the monthly ones.

(1) The time axis has one value, which is marked as a missing value.  
(2) Moreover, it has a bounds attribute which asserts the existence of a
time_bnds variable.  But no such variable exists.
There was a brief discussion on Confluence of the nature of "Where did
these files come from?  What software has to be fixed so it won't happen
again?"  But there are no answers, and we still want to deal with these
files.

That's the problem.  What can you do?

I patched cdscan to work around the second issue.  That's what gave you
the error message.  The patched cdscan is attached.
You may do ok without addressing the first issue.  I'm not sure about that.

- Jeff
```

There are also some additional tweaks to cdscan that are queued #124 which may also need some attention for a 2.2 milestone - @doutriaux1 can you think of anything else required?

durack1:
Fix #948 - @doutriaux1 following @painter1 this is an augmented version of cdscan which now works with CAM/ACME outputs - I would recommend merging this PR as a 2.2 milestone


doutriaux1:
changes look good to me. Not sure what the linux build failure is all about... merging

PR #954, opened 2015-01-08
cloning Paraview in two steps instead of one

doutriaux1:
Otherwise on RH6 /usr/bin/git is too old and can't checkout tag


doutriaux1:
@aashish24 @dlonie please approve and move tag point for 2.1.0

dlonie:
+1 for me, I agree this is the best way to handle this if older git can't handle `checkout -b [tag]`. I'll let @aashish24 approve in case I'm missing something.

aashish24:
:+1:

PR #964, opened 2015-01-14
Fix quotes

### LINKED ISSUE #963: HDF5 patch step fails with cmake-gui
The patch step for HDF5 doesn't happen on my Mac if I configure with cmake-gui, but works if I used cmake. Then build fails on the invalid `//` comment line in `h5diff.c`.

remram44:
Fixes `@...@` configure variables not being quoted, causing build failures if paths contain spaces (this **can happen** on Mac OS).

Fixes #963


aashish24:
LGTM :+1:

aashish24:
I will fix the merge conflict

PR #967, opened 2015-01-14
Fix quotes updated

aashish24:


durack1:
@aashish24 you probably want to increment README.md to show "2.1.0" as the current stable version..

PR #798, opened 2014-10-13
Fixes BUILDINSOURCE detection

remram44:
Fixes #713

Amends #736; paths are not regexps.

Re-applies #791 wrongly merged and reverted by #797; please test BEFORE merging.

(low priority)

PR #973, opened 2015-01-19
Build in source

dlonie:
Updated build in source test.

New PR is due to the old branch being in a fork (and thus not update-able by the team). #798 should be replaced by this PR.

This should go without saying, but please remain professional and respectful while discussing this patch. We're on the same team here -- I'd rather not see another issue history like #798.


dlonie:
@aashish24 @doutriaux1 @remram44 

Added tests for functionality.

aashish24:
LGTM :+1: I will just wait for Travis (as policy) and then will merge.. thanks

aashish24:
Looks like we are hitting the cmake version issue. I will push a separate fix for that.

jbeezley:
A while back I updated Girder's `.travis.yml` to update to the latest cmake.  You can install from the binaries hosted at Kitware pretty easily. See here: https://github.com/girder/girder/pull/564.

aashish24:
@jbeezley thanks. We had it for Linux but not for OSX.

PR #974, opened 2015-01-19
System python updated

aashish24:
Updated


doutriaux1:
did you test it? Last I tried it didn't work, we shouldn't approve our own branches @aashish24 will run a build now on my systems and will report any issues (if any). This shouldn't hurt the regular build so I guess it's ok.

chaosphere2112:
This PR breaks the build for OSX 10.8. I was attempting to update my dev version of UV-CDAT and ran into a bunch of build errors in the same vein; it turned out the MyProxyClient was trying to install itself into $INSTALL_PREFIX/lib/python2.7/site-packages instead of $INSTALL_PREFIX/Library/.../python2.7/site-packages, and didn't like it.

I checked out commit e114e06585c563e44ae219b733d6e3e6fb1c3abc, and the build worked fine.

doutriaux1:
@aashish24 we should either fix quickly or revert right, see, never self merge ;)

PR #975, opened 2015-01-20
Update cmake to version 3.1 on travis before building

jbeezley:
Should fix the `cmake_minimum_required` errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 language: c++
 
 before_install:
-    - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo add-apt-repository ppa:kubuntu-ppa/beta -y; sudo apt-get update; sudo apt-get install -y cmake; sudo apt-get install -y gfortran; fi
+    - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo apt-get install -y gfortran; fi
     - if [ "$TRAVIS_OS_NAME" = "linux" ]; then pushd $HOME && mkdir cmake3.1 && cd cmake3.1 && (curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-x86_64.tar.gz" | gunzip -c | tar x) && cd cmake-*/bin && export PATH="${PWD}:${PATH}"; popd; fi
     - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update ; brew outdated cmake || brew upgrade cmake ; fi
     - cmake --version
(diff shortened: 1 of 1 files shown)


doutriaux1:
@jbeezley the mac build still fails, probably because you're only looking for "linux" could you please also add th bits for Mac? Or is Mac failing for some othe reason?

jbeezley:
I'm not familiar with the build environment for mac, but I could do the same if the cmake is too old.  I believe the mac test failure is unrelated to this though.

doutriaux1:
If think mac is even more sensitive to CMake version than Linux so it might be worth doing it always no matter which OS (maybe test for Travis' CMake version first though)

jbeezley:
It looks like the osx build uses homebrew cmake at version 3.0.2.  I added a line to upgrade to 3.1 via homebrew.

.travis.yml
```diff
@@ -6,6 +6,9 @@ language: c++
 
 before_install:
     - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo add-apt-repository ppa:kubuntu-ppa/beta -y; sudo apt-get update; sudo apt-get install -y cmake; sudo apt-get install -y gfortran; fi
```
aashish24: We should then get rid of cmake install here

.travis.yml
```diff
@@ -5,7 +5,10 @@ os:
 language: c++
 
 before_install:
-    - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo add-apt-repository ppa:kubuntu-ppa/beta -y; sudo apt-get update; sudo apt-get install -y cmake; sudo apt-get install -y gfortran; fi
+    - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo apt-get install -y gfortran; fi
+    - if [ "$TRAVIS_OS_NAME" = "linux" ]; then pushd $HOME && mkdir cmake3.1 && cd cmake3.1 && (curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-x86_64.tar.gz" | gunzip -c | tar x) && cd cmake-*/bin && export PATH="${PWD}:${PATH}"; popd; fi
+    - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update ; brew outdated cmake || brew upgrade cmake ; fi
+    - cmake --version
```
aashish24: Not sure what this line would do? It won't console output I think.. (could be wrong though)

aashish24:
Other than that.. LGTM. Merging.

.travis.yml
jbeezley: Yes, you can see it in the travis output.

PR #979, opened 2015-01-23
Line duplicate points fix

chaosphere2112:
When using lines in vcs, a bug would add many duplicate points to the line object. Here's a sample:

```python
import vcs

x = vcs.init()

l = x.createline()

l.x = [[0, .5], [0, .75], [0, 1]]
l.y = [[0, .5], [0, .75], [0, 1]]

print l.x
print l.y

x.plot(l, bg=1)

print l.x
print l.y
```

which outputs

```
[[0, 0.5], [0, 0.75], [0, 1]]
[[0, 0.5], [0, 0.75], [0, 1]]
[[0, 0.5, 0.5], [0, 0.75, 0.75], [0, 1, 1]]
[[0, 0.5, 0.5], [0, 0.75, 0.75], [0, 1, 1]]
```

I spruced up the prepLine function in vcs2vtk.py, and renamed the ambiguous variables that led to the bug. I also added a test that makes sure that the X and Y points are being extended correctly.

### DIFF
--- a/Packages/vcs/Lib/vcs2vtk.py
+++ b/Packages/vcs/Lib/vcs2vtk.py
@@ -1182,29 +1182,38 @@ def prepMarker(renWin,marker,cmap=None):
   return actors
 
 def prepLine(renWin,line,cmap=None):
-  n = prepPrimitive(line)
-  if n==0:
+  number_lines = prepPrimitive(line)
+
+  if number_lines == 0:
     return
+
   actors = []
-  for i in range(n):
+
+  for i in range(number_lines):
     l = vtk.vtkLine()
     lines = vtk.vtkCellArray()
     x = line.x[i]
-    y=line.y[i]
-    c=line.color[i]
-    w=line.width[i]
-    t=line.type[i]
-    N = max(len(x),len(y))
+    y = line.y[i]
+    c = line.color[i]
+    w = line.width[i]
+    t = line.type[i]
+    number_points = max(len(x),len(y))
+
+    # Extend x or y to the length of the other by duplicating the last coord.
     for a in [x,y]:
-      while len(a)<n:
+      while len(a)<number_points:
         a.append(a[-1])
+
     pts = vtk.vtkPoints()
-    for j in range(N):
+
+    for j in range(number_points):
       pts.InsertNextPoint(x[j],y[j],0.)
-    for j in range(N-1):
+
+    for j in range(number_points - 1):
       l.GetPointIds().SetId(0,j)
       l.GetPointIds().SetId(1,j+1)
       lines.InsertNextCell(l)
+
     linesPoly = vtk.vtkPolyData()
     geo,pts=project(pts,line.projection,line.worldcoordinate)
     linesPoly.SetPoints(pts)
--- a/testing/vcs/test_vcs_line_point_extension.py
+++ b/testing/vcs/test_vcs_line_point_extension.py
@@ -1,0 +1,21 @@
+import vcs, os, sys
+
+x = vcs.init()
+line = x.createline()
+line.x = [[0, .1], [.2, .3], [.4, .5], [.6, .7]]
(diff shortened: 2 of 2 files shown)


Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -1182,29 +1182,38 @@ def prepMarker(renWin,marker,cmap=None):
...
+    c = line.color[i]
+    w = line.width[i]
+    t = line.type[i]
+    number_points = max(len(x),len(y))
+
+    # Extend x or y to the length of the other by duplicating the last coord.
     for a in [x,y]:
-      while len(a)<n:
+      while len(a)<number_points:
         a.append(a[-1])
```
dlonie: @doutriaux1 This looks like the same issue we had in 46105aafee32a0657f49859d28cf56952eb41cc4. Can you check the similar functions in this file and make sure they're treating both n and N correctly?

I think this `while len(a) < number_points: a.append(a[-1])` bit should just be `assert(len(a) == number_points)`(or similar failure) here, too, unless this is an expected feature. Otherwise it would likely just mask user mistakes and give unexpected incorrect results.

doutriaux1: Well I'm open to this, but historically it would pad it for the user with the last value.

dlonie: Gotcha. As long as it's doing what they'd expect it to!

dlonie: Oops, and I'd linked the wrong commit earlier -- I meant to point at this one: d8f725a60b5bbb1cfc26f7621db41595eec6df2d

testing/vcs/test_vcs_line_point_extension.py
```diff
@@ -0,0 +1,23 @@
...
+line.x = [[0, .1], [.2, .3], [.4, .5], [.6, .7]]
+line.y = [[0, .1], [.2, .3, .4], [.5], [.6, .7]]
+
+x.plot(line, bg=1)
+
+if line.x[0] != [0, .1]:
+  print "line.x[0] should be [0, .1]; is %s" % line.x[0]
+  sys.exit(-1)
+
+if line.x[1] != [.2, .3, .3]:
```
doutriaux1: @chaosphere2112 shouldn't that line be:
```
+if line.x[1] != [.2, .3]:
```

testing/vcs/test_vcs_line_point_extension.py
```diff
@@ -0,0 +1,23 @@
...
+  print 'line.x[1] should be [.2, .3]; is %s' % line.x[1]
```
doutriaux1: @chaosphere2112 should'nt that line say:
```
+  print 'line.x[1] should be [.2, .3,.3]; is %s' % line.x[1]
```

chaosphere2112: Ah, yeah, good catch. I'll update that.

aashish24:
is this still valid?

chaosphere2112:
Yup. Looks like it's still doing it.

```python
In [1]: l = vcs.createline()

In [2]: l.x = [[0, .5], [0]]

In [3]: l.y = [[0, .5], [0]]

In [4]: x.line(l)
Out[4]: <vcs.displayplot.Dp at 0x111a548c0>

In [5]: l.x
Out[5]: [[0, 0.5], [0, 0]]

In [6]: l.y
Out[6]: [[0, 0.5], [0, 0]]
```

doutriaux1:
@chaosphere2112 can you also please add the test to CMakeList.txt in testing/vcs ?

Thanks,

C.

chaosphere2112:
Oh, yeah, that would probably be helpful :sweat_smile:

doutriaux1:
dont' worry I did it

chaosphere2112:
Ah, thanks!

PR #985, opened 2015-01-28
fix #984

### LINKED ISSUE #984: cdutil times criteriaargclim is useless and leads to error
https://github.com/UV-CDAT/uvcdat/blob/master/Packages/cdutil/Lib/times.py#L1359
prevents to pass None as valid criteriaargclim, and with generalCriteria it leads to an error, we simply need to comment this line out

But in general I think the criteriaarg for clim is not needed and can probably be completely removed sinceI don't think anybody ever came (or will come up) with a personalized criteria.

doutriaux1:
criteriaargclim was messing up criteriaarg and useless
commented out, probably should take all code out, but in case of 'custom' criteria func it might be usefull
so letting it in

### DIFF
--- a/Packages/cdutil/Lib/times.py
+++ b/Packages/cdutil/Lib/times.py
@@ -1356,7 +1356,7 @@ class Seasons(ASeason):
         '''
         cdat_info.pingPCMDIdb("cdat","cdutil.times.Seasons.climatology -%s-" % self.seasons)
         u=self.month_fix(slab)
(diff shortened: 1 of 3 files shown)


doutriaux1:
@dlonie @aashish24 @painter1 this is a tiny fix to a HUGE problem for diags test suite please review and merge as soon as possible we need this for ACME project.
@painter1 in the mean time you can comment out the line on your builds.

dlonie:
LGTM

PR #987, opened 2015-01-28
Create an option to enable vtkweb

jbeezley:
It defaults to off so nothing should blow up.

### DIFF
--- a/CMake/cdat_modules/vtk_external.cmake
+++ b/CMake/cdat_modules/vtk_external.cmake
@@ -73,6 +73,12 @@ if(CDAT_BUILD_OFFSCREEN)
   )
 endif()
 
+if(CDAT_BUILD_VTKWEB)
+  list(APPEND vtk_build_args
+    "-DVTK_Group_Web:BOOL=ON"
+  )
+endif()
+
 set(_vtk_module_options)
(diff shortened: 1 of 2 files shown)


CMake/cdat_modules/vtk_external.cmake
```diff
@@ -73,6 +73,12 @@ if(CDAT_BUILD_OFFSCREEN)
   )
 endif()
 
+if(CDAT_BUILD_VTKWEB)
```
aashish24: How about CDAT_ENABLE_WEB? Or CDAT_BUILD_WEB?

jbeezley: `CDAT_BUILD_*` is consistent with the other options.  To me, `VTKWEB` is more explicit unless there is something else that `WEB` will enable, but I don't have a strong opinion.

doutriaux1: I like that: CDAT_BUILD_VTKWEB

aashish24: @doutriaux1 we don't want to expose vtkweb. Web option in the future will enable more than just vtkWeb.

aashish24:
LGTM :+1:

PR #990, opened 2015-01-29
Merge remote-tracking branch 'origin/fix-quotes-updated' into release

doutriaux1:


doutriaux1:
@aashish24 @remram44 that seems to be doing the trick can you approve? I'm all against self approving.

Issue #991, opened 2015-01-29
mac build broken


doutriaux1:
system python merge broke build on macs ( @aashish24 please do not do that again :smile: )
PYTHONPATH isn't set correctly then pip/easy_install complains about PYTHONPATH not containing directory where we try to install packages.

jbeezley:
See #988, I think that takes care of it.

aashish24:
@doutriaux1 we had system python branch sitting there for weeks (we had to do it :cry: ). Hopefully the fix pushed by @jbeezley will take care of that issue. Also, note that it seems to be broken beforehand (looking at the changes).

ThomasMaxwell:
I just merged my 2.1.1 contributions branch into master and tried building under OSX 10.8.5.   I encountered the following build error.   Charles, is this the error you were referring to in #991?   Is there a fix?

TEST FAILED:
  /Developer/Projects/EclipseWorkspace/uvcdat/master/build1/install/lib/python2.7/site-packages/
  does NOT support .pth files
  error: bad install directory or PYTHONPATH

  You are attempting to install a package to a directory that is not on PYTHONPATH and which Python does not read ".pth" files from.  The  installation directory you specified (via --install-dir, --prefix, or  the distutils default setting) was:

      /Developer/Projects/EclipseWorkspace/uvcdat/master/build1/install/lib/python2.7/site-packages/

  and your PYTHONPATH environment variable currently contains:

 '/Developer/Projects/EclipseWorkspace/uvcdat/master/build1/install/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages'

remram44:
I'm also seeing this build failure. [Same log as Travis](https://travis-ci.org/UV-CDAT/uvcdat/jobs/48775517#L4294):
```
  You are attempting to install a package to a directory that is not
  on PYTHONPATH and which Python does not read ".pth" files from.  The
  installation directory you specified (via --install-dir, --prefix, or
  the distutils default setting) was:

      /Users/remirampin/Documents/programming/uvcdat/builds/4f89f6eb442faf03979a4e0706fc5fd4d24aa352/install/lib/python2.7/site-packages/

  and your PYTHONPATH environment variable currently contains:

      '/Users/remirampin/Documents/programming/uvcdat/builds/4f89f6eb442faf03979a4e0706fc5fd4d24aa352/install/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages'
```

[...]

PR #998, opened 2015-01-30
Need to escape an additional % in install.py

jbeezley:
This [commit](https://github.com/UV-CDAT/uvcdat/commit/4bc10968e29742f0714dff26971c893713fe5fa8#diff-364b4e9b07d47ae67379edb8858a5b19) missed a `%` that needs to be escaped resulting in the error mentioned [here](https://github.com/UV-CDAT/uvcdat/issues/991#issuecomment-72233051).  I'm still building locally, so I can't yet verify if this works.


jbeezley:
Looks like that fixed my local build.  Also, `vcs` imports in the built environment.

remram44:
Built on my machine.

PR #1013, opened 2015-02-06
Export text as paths in vector exports.

### LINKED ISSUE #503: text rotation in VTK different from vcs way
If I "left justify" a text, and then rotate it, the "justify" seems to be applied to the rotated text.
Trying to show this here, I have a vertical centered and left justified text.

First putting 3 points on the same "y" location I would expect (with the letter rotated 90 degrees)
```
A     A
B     B
C  A  C
D  B  D
```
But in fact I get
```
A     A
B  A  B
C  B  C
D     D
```

dlonie:
The text alignment flags aren't handled consistently across GL2PS
backends (e.g. PDF has no concept of alignment), so exporting the text
as a series of Bezier curves will ensure that the resulting export is
consistent, regardless of format and/or viewer.

This fixes #503. Note that I had to fix an issue with VTK for this to handle the custom fonts properly -- make sure UV-CDAT/VTK@9760973f6acbfc64a3a29ffecda30ebd6675310a is in your VTK tree when testing (VTK/uvcdat-master has already been updated).

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1338,6 +1338,7 @@ class VTKVCSBackend(object):
     gl.SetInput(self.renWin)
     gl.SetCompress(0) # Do not compress
     gl.SetFilePrefix(".".join(file.split(".")[:-1]))
+    gl.TextAsPathOn()
     if output_type=="svg":
         gl.SetFileFormatToSVG()
     elif output_type == "ps":


aashish24:
LGTM @doutriaux1 can you get this merged soon?

aashish24:
@doutriaux1 we really need this fix. Can we get it merged?

doutriaux1:
@aashish24 I trust it didn't break any test case ;) otherwise please push the baseline as soon as possible.

aashish24:
thanks @doutriaux1 :+1:

PR #1014, opened 2015-02-06
Update background color

chaosphere2112:
If you update canvas.backgroundcolor after rendering and do canvas.update(), the canvas' background color does not update. Added tests for that, and fixed the bug. I'm up for moving the fix to a different place, but the backend.clear() function was the most convenient location (since there was already an if statement that separated out the specific renderer I was looking to update, and gets called during every update()). Incoming PR for uvcdat-testdata to provide the images as well.


aashish24:
@chaosphere2112 can we have bit more interest test?

doutriaux1:
@aashish24 boringcan be good sometimes! @chaosphere2112 I do not like your yellow, but I will approve anyway.

Issue #517, opened 2014-07-25
pdf files are HUGE


doutriaux1:
@aashish24 there must be a way for VTK to produce smaller pdfs no?

aashish24:
Re-opened just in case.

dlonie:
Unassigning myself -- can't do much about this without an example ;)

aashish24:
Right.. assigning it to Charles for now as we need to have an example.

doutriaux1:
all right data is at:
https://www.dropbox.com/home/HUGE%20PS
how to reproduce:
notice the time (over 9 minutes) .... and the errors messages
reassignnig to @dlonie now that he has something to work with.
Thanks
```
In [1]: import cdms2,vcs

In [2]: x=vcs.init()

In [3]: f=cdms2.open("pr_TRMMtS2_Jul99-08.nc")

In [4]: s=f("prMAXt")

In [5]: x.plot(s)
Out[5]: <vcs.displayplot.Dp at 0x7f90791d39b0>

In [6]: import time

In [7]: a=time.time();x.posts
x.postscript      x.postscript_old  

In [7]: a=time.time();x.postscript("test_slow_and_big");print time.time()-a
/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/VTKPlots.py:1371: UserWarning: the right_margin keyword for postscript has been deprecated in 2.0 and is being ignored
  warnings.warn("the right_margin keyword for postscript has been deprecated in 2.0 and is being ignored")
/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/VTKPlots.py:1373: UserWarning: the left_margin keyword for postscript has been deprecated in 2.0 and is being ignored
  warnings.warn("the left_margin keyword for postscript has been deprecated in 2.0 and is being ignored")
/lgm/uvcdat/2015-01-26/lib/python2.7/site-packages/vcs/VTKPlots.py:1375: UserWarning: the top_margin keyword for postscript has been deprecated in 2.0 and is being ignored
  warnings.warn("the top_margin keyword for postscript has been deprecated in 2.0 and is being ignored")
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
[...]

PR #1016, opened 2015-02-09
Increase the initial GL2PS feedback buffer size.

dlonie:
To prevent having to resize and rerender multiple times when a very large
amount of data is pushed through (see #517), start with a 50MB buffer.

This buffer size is sufficient to store the 1M triangles used in the #517
example.

@doutriaux1 @aashish24

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1340,7 +1340,12 @@ class VTKVCSBackend(object):
     # vertices produced by OpenGL. If you start seeing a lot of warnings:
     # GL2PS info: OpenGL feedback buffer overflow
     # increase it to save some time.
-    gl.SetBufferSize(50*1024*1024) # 50MB
+    # ParaView lags so we need a try/except around this
+    # in case it is a ParaView build
+    try:
+      gl.SetBufferSize(50*1024*1024) # 50MB
+    except:
+      pass
 
     # Since the vcs layer stacks renderers to manually order primitives, sorting
     # is not needed and will only slow things down and introduce artifacts.
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1342,6 +1342,10 @@ class VTKVCSBackend(object):
     # increase it to save some time.
     gl.SetBufferSize(50*1024*1024) # 50MB
 
+    # Since the vcs layer stacks renderers to manually order primitives, sorting
+    # is not needed and will only slow things down and introduce artifacts.
+    gl.SetSortToOff()
+
     gl.SetInput(self.renWin)
     gl.SetCompress(0) # Do not compress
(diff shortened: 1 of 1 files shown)


aashish24:
LGTM but lets @doutriaux1 approve it as well.

doutriaux1:
@dlonie 
I get this, using a full build (i.e Paraview vs VTK) could that be the issue?
```
/usr/local/uvcdat/2015-02-09/lib/python2.7/site-packages/vcs/VTKPlots.pyc in vectorGraphics(self, output_type, file, width, height, units)
   1341     # GL2PS info: OpenGL feedback buffer overflow
   1342     # increase it to save some time.
-> 1343     gl.SetBufferSize(50*1024*1024) # 50MB
   1344 
   1345     # Since the vcs layer stacks renderers to manually order primitives, sorting

AttributeError: SetBufferSize
```

doutriaux1:
```
In [10]: v.GetVTKVersion()
Out[10]: '6.2.0'
```

dlonie:
Yes -- this patch is only in the VTK uvcdat-master.

@aashish24, what is our strategy for syncing paraview and VTK?

aashish24:
@dlonie I have updated the VTK in UV-CDAT/VTK. I just pushed the latest VTK with one extra commit on top.

aashish24:
@doutriaux1 make sure that you force rebuild VTK.

dlonie:
@aashish24 What about FULL builds with ParaView, though? How do we get these changes in there?

doutriaux1:
@aashish24 @dlonie is right the issue is that these change seem to not be present when building ParaView. Can we have UV-CDAT's Paraview point to the same uvcdat-master tag point of UV-CDAT's VTK

aashish24:
@doutriaux1 yes but ParaView is mostly behind. I would have to check on the SHA its using.

Do you want to wait for that? I think most of us are using VTK so I prefer we merge it right now but I don't have strong feelings about it.

doutriaux1:
most of you are using default, but most of my users come down to my office if i accidentally build the default one. And Curt who reported this bug first is one of them. Why can't we update ParaView's VTK?

doutriaux1:
I saw your email. Building from scratch now, will get back to you.

doutriaux1:
@aashish24 ok now the whole ParaView is broken... I guess I will put a try/xcept around this so it works even with PAraView.
Can you please revert ParaView tag point. Thanks,
```
From git://vtk.org/VTK
   cdb450f..e4a03fa  master     -> origin/master
fatal: reference is not a tree: aa7eeca6abbb25d4a76e81028b10de24d0e6f6a8
Unable to checkout 'aa7eeca6abbb25d4a76e81028b10de24d0e6f6a8' in submodule path 'VTK'
```
I guess my question is why is it checking out VTK at VTK.org and not from us?

aashish24:
@doutriaux1 sorry, I had one commit there left by mistake. Try it now. We cannot use our VTK since the URL would be diffirent and then also we have to make sure that it is compabille with ParaView.

PR #1025, opened 2015-02-10
Issue1023 durack1 cython0p21p2 update

durack1:
@doutriaux1 @aashish24 can this be merged into master once the travis build has passed? I presume that Cython is only built under "full" build conditions?


durack1:
@doutriaux1 it seems to have passed but not sure travis actually builds cython...?

aashish24:
@durack1 right, Cython is not built in default mode (which travis uses)

durack1:
@aashish24 would it be a good idea to convert travis to build full? It kinda seems silly to have tests running when they're missing half the modules which are built under a full installation..

doutriaux1:
@aashish24 can we have travis build both?

aashish24:
@doutriaux1 the problem is travis limits the build time to 50  minutes on a two core machine. Last time I tried a full build was more than that. We can try again.

doutriaux1:
yes it is in the limits... but we could could turn off jpeg/libxml2/libxslt/zlib/readline these are on all system and can take quite some time. Also on Travis we can turn off FFMPEG since it is not used by test suite.

aashish24:
@doutriaux1 the proper way would be to modify cmake to use system ones. Let's plan for it after 2.2

durack1:
@doutriaux1 @aashish24 this discussion has wandered far away from the PR which was simply trying to update Cython.. Do you want me to create a new issue detailing the thread above - that is more effectively using travis for full build testing?

durack1:
@doutriaux1 is this good to merge? I note that lxml, pyzmq, vistrails and pytables all have Cython listed as a dependency.. Are any current tests covering this package then, does 0.21.2 work for these dependencies?

PR #1041, opened 2015-02-17
fix #976

### LINKED ISSUE #976: Longitude and latitude labels
The 3 types of 2d plots have different default labels. I think they should be the same. Also, itwould be more useful if the latitude labels went from -90 to 90.

The longitudes should start at 0 not 10E then with 30 degree increments 180 would be labeled - It is necessary to know the international date line when the data is wrapped with the Pacific Ocean as the center. 
If nothing else, all the graphics method should be the same. See the attached figure.
latitude_longitude_labels

doutriaux1:
bxofill wasn't using going through path of resetting default lat/lon
lon30 was not quite right (starting at -170)
because it is also in initial.attributes and may be used by existing scripts
I decided to create Lon30
same for lat20, create Lat20 which includes 90N/90S

@aashish24 @dlonie @chaosphere2112 don't forget new baselines s well


doutriaux1:
goes with https://github.com/UV-CDAT/uvcdat-testdata/pull/20

dlonie:
LGTM, but I'll let someone more familiar with usage approve, especially for the changes to case (lon20 vs Lon20, etc).

doutriaux1:
@williams13 @potter2 can you please confirm that the new baselines at: https://github.com/UV-CDAT/uvcdat-testdata/pull/20 look like what you want so we can merge. Thanks.

doutriaux1:
@aashish24 I updated the baselines to reflect @dlonie branch for isolines labels

aashish24:
LGTM :+1:

Issue #1045, opened 2015-02-18
cdscan issue - invalid time coords


durack1:
The behavior when encountering problematic time coords seems to have changed with recent tweaks to cdscan (#949):

```
[durack1@crunchy ~]$ /usr/local/uvcdat/2014-10-02/bin/cdscan -x cmip5.HadGEM2-ES.piControl.r1i1p1.mo.atm.Amon.hus.ver-v20130114.latestX.xml.2014-10-02 /cmip5_css02/scratch/cmip5/output1/MOHC/HadGEM2-ES/piControl/mon/atmos/Amon/r1i1p1/v20130114/hus/*.nc
Finding common directory ...
Common directory: /cmip5_css02/scratch/cmip5/output1/MOHC/HadGEM2-ES/piControl/mon/atmos/Amon/r1i1p1/v20130114/hus/
Scanning files ...
/cmip5_css02/scratch/cmip5/output1/MOHC/HadGEM2-ES/piControl/mon/atmos/Amon/r1i1p1/v20130114/hus/hus_Amon_HadGEM2-ES_piControl_r1i1p1_185912-188411.nc
Setting reference time units to days since 1859-12-01
...
Warning, file hus_Amon_HadGEM2-ES_piControl_r1i1p1_217312-219811.nc, dimension time contains values in file hus_Amon_HadGEM2-ES_piControl_r1i1p1_185912-188411.nc
Traceback (most recent call last):
  File "/usr/local/uvcdat/2014-10-02/bin/cdscan", line 1657, in <module>
    main(sys.argv)
  File "/usr/local/uvcdat/2014-10-02/bin/cdscan", line 1364, in main
    combineKeys(timeproj, timedict, timeIsLinear, referenceDelta)
  File "/usr/local/uvcdat/2014-10-02/bin/cdscan", line 316, in combineKeys
    raise RuntimeError, 'Dimension %s in files %s [len(%s)=%d], %s [len(%s)=%d], is inconsistent'%(name, prevpath, name, (jind-iind+1), path, name, len(values))
RuntimeError: Dimension time in files hus_Amon_HadGEM2-ES_piControl_r1i1p1_217312-219811.nc [len(time)=1], hus_Amon_HadGEM2-ES_piControl_r1i1p1_185912-188411.nc [len(time)=300], is inconsistent

[...]

PR #1047, opened 2015-02-18
1. In the section involving 'allaxesdummy', where a dummy variable is cr...

painter1:
...eated to ensure the inclusion of all axes,

now only small axes (<100) are forced to be included.
2. In a call of disambig, somehow the parentheses in var.size() had disappeared.  I restored them.


durack1:
Fix #1045

doutriaux1:
lgtm

Issue #1048, opened 2015-02-18
average() indentation error


painter1:
In cdutil/times.py, the function average(), the block beginning "if len(sub)>1:" is indented too far.  The effect is to repeat a calculation more than needed.  This has major performance implications.

doutriaux1:
@painter1 can you push a PR for this since you already have it on your computer? Thanks!

painter1:
Yes, I'll do a PR but want to test some more first.

doutriaux1:
That's awesome, even better would be to include the test in the test suite. If you don't remember how do it, just put the tests in testing/genutil  and I will integrate them.

painter1:
What I did was to run some diagnostics which print out scalars.  I happen to have some old values in my notes so I could eyeball them and see that they aren't changed.  It will be another job to turn this into an automated test.  Note that a few gigabytes of data are required.  Here's what I ran:
diags --path ~/metrics_data/cam35_data/ --path2 /Users/painter1/metrics_data/obs_data_5.6/  --package AMWG --set 1 --seasons ANN

durack1:
@painter1 out of curiosity what's the speed up by removing this unintentional loop?

painter1:
I'm only profiling on Rhea (at Oak Ridge) which is abysmally slow for reasons which I haven't identified, but must be about the OS and file system.  I'm not sure that you will notice the difference on any other machine, maybe you will.  Anyway, on my test which is probably typical, the core average calculation was repeated 12 times before and is being done once now.  On Rhea, the overall speedup is 3 times on the largish test data I've been using.

painter1:
The test I've been profiling is climatology.py in the UV-CDAT diagnostics.

doutriaux1:
fixed by #1050

PR #1050, opened 2015-02-18
Reduce indentation level of the block beginning "if len(sub)>1:" in aver...

painter1:
...ager().

This will dramatically improve performance because some major calculations are done only once now.
This fixes issue #1048.


doutriaux1:
@dlonie would have probably liked to see these very old print statements removed ( :wink: ) but that's ok by me and in the interest of time I'm merging this.

PR #1061, opened 2015-02-21
Issue1034 durack1 add h5py

aashish24:


CMake/cdat_modules/h5py_pkg.cmake
```diff
@@ -0,0 +1,18 @@
...
+set( H5PY_URL ${LLNL_URL} )
+set( H5PY_GZ h5py-${H5PY_VERSION}.tar.gz )
+set( H5PY_MD5 80c9a94ae31f84885cc2ebe1323d6758)
+
+set (nm H5PY)
+string(TOUPPER ${nm} uc_nm)
+set(${uc_nm}_VERSION ${${nm}_MAJOR}.${${nm}_MINOR}.${${nm}_PATCH})
+set(H5PY_SOURCE ${H5PY_URL}/${H5PY_GZ})
+
+if (CDAT_BUILD_ALL)
```
doutriaux1: @aashish24 I guess that works too ;) Was there no way to get it going with the add_Cdat_package_dependent?

Issue #1063, opened 2015-02-21
test failing because mean is different 


aashish24:
https://open.cdash.org/testDetails.php?test=315268558&build=3704218

doutriaux1:
that one is odd. I reuploaded a new value for mean because on my mac it seems different than on my ubunutu! LEt me retest that one.

doutriaux1:
looking into this one. Turns out the actual data sent to plot is slightly different between mac and linux...
the test code does:
```
  nm_xtra="_amplitude"
  U=numpy.cos(lons[:])
  V=numpy.sin(lats[:])
  A=MV2.array(V[:,numpy.newaxis]*U[numpy.newaxis,:])
  A.setAxis(0,lats)
  A.setAxis(1,lons)
  u*=A  
  v*=A  
```
That could explain the precision difference
diff plot:
![diff](https://cloud.githubusercontent.com/assets/2781425/6363094/33a0de20-bc49-11e4-9f71-b4faae624e30.png)

PR #1076, opened 2015-02-25
Adding 3 to amplitude so that os roundup diffs do not show in mean

doutriaux1:
@aashish24 os round up error where causing this see #1063 adding 3 to amplitude eliminates the issue
@chaosphere2112 @dlonie @aashish24 please review and merge


doutriaux1:
don't forget: https://github.com/UV-CDAT/uvcdat-testdata/pull/22

aashish24:
LGTM. Although you may want to consider doing some check in VCS itself on precision so that it can be handled gracefully.

PR #1089, opened 2015-02-26
Issue423 durack1 fix vacumm dependencies

durack1:
@doutriaux1 @aashish24 this should fix all the dependencies for vacumm - I'll just double check all these and will repost a comment here once I've confirmed all is ok - after confirmation I'd appreciate if you can merge..


durack1:
@doutriaux1 @aashish24 ok this looks fine (just double checked the dependencies for configobj and six and confirmed they're ok) - so please merge once you've reviewed.

@doutriaux1 I'm not sure that six has a dependency on pip - which is what was already in the file six_deps.cmake - it's setup.py lists no dependencies..

aashish24:
Okay, I will review it in next few hours and will get it merged..

durack1:
@aashish24 great, as noted above it could be a good idea to confirm the dependencies for six, I think pip can be removed..

durack1:
@aashish24 @doutriaux1 let's wait with this PR until we get an all clear from @stefraynaud, apparently pytz was needed in the most recent vacumm version.

durack1:
@stefraynaud, so we currently have:
`set(vacumm_deps ${python_pkg} ${numpy_pkg} ${scipy_pkg} ${matplotlib_pkg} ${basemap_pkg} ${configobj_pkg})` as dependencies, if we add `pytz` will that be enough to get reasonable functionality from the package - the sphinx dependencies to compile documentation I don't think is needed and seawater v3.3.2 is already in..

In particular it's not clear to me whether paramiko, xlw or PIL are required?

stefraynaud:
@durack1 I will commit a 2.5.1 version tomorrow that will remove the dependency to pytz.
So, as you said, `set(vacumm_deps ${python_pkg} ${numpy_pkg} ${scipy_pkg} ${matplotlib_pkg} ${basemap_pkg} ${configobj_pkg})` is sufficient to have access to most of the library.

paramiko, xlw or PIL are not required.

durack1:
@stefraynaud great thanks for clarifying, I'll download your 2.5.1 sources and update the tar ball on the server when it's available and then this PR should be good to go..

doutriaux1:
@durack1 since we provide sphinx why not add it to the dep list? It will only change the order in which it is built and should give us a more complete version in case @stefraynaud checks for it at install time. Same for seawater

durack1:
@doutriaux1 the additional packages are not included in the existing ALL builds: `sphinxcontrib-cheeseshop 0.2, sphinxcontrib-ansi 0.6, sphinxcontrib-programoutput 0.8`

doutriaux1:
I see

stefraynaud:
@durack1 Vacumm 2.5.1 is released.
@doutriaux1, there are currently no check at install time, despite numpy is needed by `setup.py`. Sphinx contribs are needed only to build the documentation.

durack1:
@stefraynaud I'll pull down the upddate tarball and revise the PR today, hopefully this might sneak under the 2.2 milestone..

doutriaux1:
@durack1 I will try to out it in

durack1:
@doutriaux1 it's updated and good to go!

CMake/cdat_modules/vacumm_deps.cmake
```diff
@@ -0,0 +1 @@
+set(vacumm_deps ${python_pkg} ${numpy_pkg} ${scipy_pkg} ${matplotlib_pkg} ${basemap_pkg} ${configobj_pkg})
```
doutriaux1: no seawater?

CMake/cdat_modules/configobj_deps.cmake
```diff
@@ -0,0 +1 @@
+set(configobj_deps ${python_pkg} ${six_pkg})
```
doutriaux1: no numpy? no setuptools? if there's a setup.py it probably needs setuptools (distutils)

CMake/cdat_modules/vacumm_deps.cmake
doutriaux1: no numpy? no setuptools? if there's a setup.py it probably needs setuptools (distutils)

CMake/cdat_modules/configobj_deps.cmake
durack1: The only dependency listed in `six` https://github.com/DiffSK/configobj/blob/master/setup.py#L97
I'm uncertain if it requires setuptools (does setup.py expect this?)

CMake/cdat_modules/vacumm_deps.cmake
durack1: Seawater isn't a hard dependency - only for advanced diags: http://relay.actimar.fr/~raynaud/vacumm/user.install.prereq.html

durack1:
@doutriaux1 setuptools added to both as dependency - good catch, thanks

durack1:
@doutriaux1 I just checked and it seems that setuptools is available in the default python redhat install, same with OS X - so not sure it needs to be built..

Mind you, it is a pretty old version:
```
[durack1@oceanonly ~]$ which python
/usr/bin/python
[durack1@oceanonly ~]$ python
Python 2.6.6 (r266:84292, Nov 21 2013, 10:50:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> setuptools.__version__
'0.6'
[durack1m1:~] durack1% python
Python 2.7.5 (default, Mar  9 2014, 22:15:05) 
[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> setuptools.__version__
'0.6c12'
```

PR #1098, opened 2015-03-02
Fixes autot resizing when configurator instantiated

chaosphere2112:
There's an issue when you plot using a ratio– as you resize the canvas, the plot would only shrink, and never grow. This fixes that; it turned out that the logic that was used to scale the template's data section to keep it in aspect ratio was breaking because I was substituting in a new template (which enables some functionality like resetting the template to the way it was before you moved things around). @doutriaux1 and I found this bug, and we've addressed it in the below commit, which fixes another bug (I was constantly generating new substitute templates, but now will only generate one once per base template).


doutriaux1:
works for my 10.10

Issue #1096, opened 2015-03-02
No usch text combined


painter1:
When I use the GUI to make a diagnostics plot it looks good - until I move the mouse over it.  Then something raises the following exception:

Exception: No usch text combined: __texttable_733759201101579:::__textorientation_833871807321088
Traceback (most recent call last):
  File "/Users/painter1/bin/uvcdat-2015.03.02/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/configurator.py", line 269, in hover
    obj = self.in_display_plot(point, display)
  File "/Users/painter1/bin/uvcdat-2015.03.02/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/configurator.py", line 384, in in_display_plot
    tc = vcs.gettextcombined(dp.g_name)
  File "/Users/painter1/bin/uvcdat-2015.03.02/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/manageElements.py", line 1613, in gettextcombined
    raise Exception,"No usch text combined: %s:::%s" % (Tt_name_src,To_name_src)

Thereafter, if I click on another unrelated window (e.g. for a shell) the plots disappear.

painter1:
This is with a new UV-CDAT built less than an hour ago, off a pull of UV-CDAT master and VisTrails uvcdat-master, and a pull of the metrics dvel branch.

painter1:
After I redrew the plots, they wouldn't disappear any more.

doutriaux1:
@chaosphere2112 could it be linked to configurator?

chaosphere2112:
@doutriaux1 Looks like it definitely is; @painter1, how about I swing by and you show me how to reproduce?

chaosphere2112:
Just submitted a pull request to fix this issue, should get merged soon.

chaosphere2112:
Should be fixed in latest master.

PR #1107, opened 2015-03-03
Properly remove template attribute displayplots (fix for #1096)

chaosphere2112:
Got rid of some duplication, and moved the code to the correct location. Fixed issue #1096.


chaosphere2112:
@doutriaux1 @aashish24

PR #1113, opened 2015-03-05
VCS2D Animation while Interacting

chaosphere2112:
Big improvements to the animation controls during interaction; play works a *lot* better, can step a frame forward or backward, and added the ability to save out the animation and choose the framerate. Had to tweak a few things in the animation classes to make sure they worked with the changes, but I don't think I broke anything by doing so.


doutriaux1:
hum.. just tried it from command line ubuntu (with merged master in)
configure button are now behind plots, so hard to get to
clicking (behind plot) on run, gen a lot of blinking.
pressing Q won't work anymore...
eventually (when pngs are done i think) flickering goes away , buttons stay on top, but pressing stop button does not work
Don't see a slider either. 
Am i using a bad version?

chaosphere2112:
Nope, looks like it's doing that on mine as well. I'll keep working on this.

chaosphere2112:
So, I guess you shouldn't just keep spawning repeating timers in your timer event listener. Whatever. That would be why the Q key and the Stop Animation button didn't work– I was flooding the event loop.

Still have the UI flashing; it's because of the png generation. I'll keep working on that.

chaosphere2112:
@doutriaux1 I can't think of any great way to *not* flash the UI during animation, given that we want to generate the PNGs. I guess we could leave the UI in, and then when you save, rebuild the PNGs and strip out the UI for those? Dunno, let's chat on Monday about it.

doutriaux1:
I don't remember it flashing that way before. Before I was not using pngs until they were all ready so there was not flashing on first pass at all, second pass was super fast. Why do you need pngs on first pass? I think the issue is that before widgets come in it was barely noticeable and now the redraw time gets a bit longer and it is noticeable.

chaosphere2112:
Sorry, miscommunication; the animation doesn't flicker at all anymore (since I fixed the timers spawning timers spawning timers issue), and runs fine. The widgets flash off and then on as the PNG is rendered.

chaosphere2112:
@doutriaux1 Also, I think I've got an idea for how to get around this issue. I'm going to experiment to see if it works, and we can discuss if it seems like a good idea or not. Are you in this afternoon, or are you out today?

doutriaux1:
I'm in but from home. Want to setup a gotomeeting at some point?

chaosphere2112:
Got a sec to chat on the phone? I can explain my idea and see if you think it's crazy or not.

chaosphere2112:
@doutriaux1 Alright, I got offscreen rendering working for animation. I had to add some logic so run() would do it straight in the main window; rendering offscreen on a non-main thread was making the objective-c runtime throw a hissy fit. Mind giving this branch a shot?

doutriaux1:
almost there! Works like a charm from command line. BUT (theres always a but) in GUI I can't get the widget to show up (even clicking on edit pen). Am I missing something?

chaosphere2112:
@doutriaux1 I haven't done the configure pen integration yet; I wrote the code to handle enabling and disabling the UI at the canvas level in another branch, but had not yet tied in with the configure pen in VisTrails. I'll have no problem knocking that out Monday morning. Presumably this should be easy to fix then as well.

doutriaux1:
Should I merge that and you start a new branch? Or you keep working on this one?

chaosphere2112:
I'll just do it in this one. I'll get that going right now.

chaosphere2112:
OK, I fixed the UI not showing up in the GUI. Some of the buttons are still showing up behind the text. I made it so my stuff only happens while the configure menu is open, though you'll have to use this branch:

https://github.com/chaosphere2112/VisTrails/tree/configure_button_2d_interactions

I'll submit a PR to VisTrails when we have that UI nailed down.

chaosphere2112:
I'm getting a segfault while resizing UV-CDAT with the configure menu open; working on sorting that out now.

doutriaux1:
might be related to issue sI'm seeing on: #1135

chaosphere2112:
OK @doutriaux1, I'm pretty sure I fixed the issue you were having with #1135. Also, I implemented what we discussed with regards to the extra render calls coming from `renderTemplate`, by adding a "fromtemplate" argument to all of the strings being rendered by the template, though now I'm getting this warning message:

```
warnings.warn('Unrecognized vcs plot keyword: %s, assuming backend (%s) keyword'%(keyarg,self.backend.type))
/Users/samfries/uvcdat-master/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.py:3501: UserWarning: Unrecognized vcs plot keyword: fromtemplate, assuming backend (vtk) keyword
```

I assume there's an easy way to quash that?

chaosphere2112:
Also, @doutriaux1, I can't test the resizing in uv-cdat on my local machine at home; my app complains about not having CDMS installed still (Yosemite build).

aashish24:
@chaosphere2112 I will review this branch carefully. Please don't merge until then.

chaosphere2112:
@aashish24 I've just tested it in the GUI, and it's still a little laggy on resizing. I've got some more work to do there before a merge happens.

doutriaux1:
@chaosphere2112 yes, please add any new keyword there:
https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/Canvas.py#L2423

chaosphere2112:
@doutriaux1 This should work when used with my [VisTrails](https://github.com/chaosphere2112/VisTrails/tree/configure_button_2d_interactions) branch. I'll go ahead and submit the PR for that right now. The one thing that's still an issue is that template changes aren't being persisted in any useful fashion right now; I should be able to knock that out tomorrow, in a separate pull request (I think this one has gotten bloated enough :wink:).

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -7,15 +7,92 @@
...
+    while ren:
+      i+=1
+      actors = ren.GetActors()
+      actors.InitTraversal()
+      actor = actors.GetNextItem()
+      j=0
+      while actor:
+        j+=1
+        m = actor.GetMapper()
+        m.Update()
```
aashish24: Why we are manually updating the mapper?

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -7,15 +7,92 @@
...
+        actor=actors.GetNextItem()
```
aashish24: Every frame updating like this is not ideal in a pipeline based system.

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -7,15 +7,92 @@
...
+      ren=renderers.GetNextItem()
+
+    self.hidden_window.Render()
```
aashish24: can we call it offscreen_window please?

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -41,48 +138,85 @@ def __init__(self,vcs_self):
...
+                self.create_thread.hidden_window.AddRenderer(ren)
+            else:
+                self.create_thread.bg_ren.SetBackground(*ren.GetBackground())
+            ren = renderers.GetNextItem()
+        be.showGUI()
+
+        self.create_thread.hidden_window.SetSize(*be.renWin.GetSize())
+
+    def reclaim_renderers(self):
+        if not self.cleared:
```
aashish24: Could we please add some documentation?

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -41,48 +138,85 @@ def __init__(self,vcs_self):
...
+
+        be = self.vcs_self.backend
+
+        if be.renWin is None: #Nothing to clear
+            return
+
+        be.hideGUI()
+        renderers = be.renWin.GetRenderers()
+        renderers.InitTraversal()
+        ren = renderers.GetNextItem()
```
aashish24: I have to think more about it but this could lead to problems if the renderer knows about the context from the previous render window.

aashish24:
@chaosphere2112 please have a look at the comments. 

also 

@chaosphere2112 could please adopt the practice of adding some documentation to each method we add? Its just makes the code review easier and improves the code quality. While developing new code its much easier to add high level description of the method or class. I am going to push others for the same.

Packages/vcs/Lib/VTKAnimate.py
chaosphere2112: I just copied the `update_input` function's `update` section from [VTKPlots.py](https://github.com/UV-CDAT/uvcdat/blob/6a25e00622c953ecc4dd8ffdd8f177f0daf3b350/Packages/vcs/Lib/VTKPlots.py#L1845-L1863) and applied it to my background render window. I'm more than happy to swap out for another way of doing this, I was just trying to do a minimum of inventing new logic.

Packages/vcs/Lib/VTKAnimate.py
chaosphere2112: Sure, I can make that change.

Packages/vcs/Lib/VTKAnimate.py
chaosphere2112: Sure, I can do that.

chaosphere2112:
@aashish24 I've addressed your concerns; mind taking another pass?

aashish24:
@chaosphere2112 thanks. one thing we don't understand. Why you have two render windows? When you move actors from one to another, it has to destroy the display list and the VBO's anyways from the first one?

chaosphere2112:
@aashish24 The issue that I'm solving is that the GUI would flash during each frame of the animation. It did this because we export out the PNG file the first time the frame is rendered, and to do this, we had to remove the UI for a second, render the window without the UI for the PNG, and then bring the GUI back. This made it basically impossible to stop the animation (due to the button constantly disappearing) until after an entire pass had happened, and we could just load images instead of updating the actors. It was also a really bad experience for users. By moving the renderers to a BG render window, we can render the updated actors to a PNG without having to do anything to the UI. It also gives the option of making the PNGs faster than we are displaying them, though we don't at this time.

chaosphere2112:
Hm, looks like the animation will error when run from a script, instead of being launched via interactive console, UI, or GUI.

doutriaux1:
what kind of error? I'm using this to test:
```
import sys,cdms2,vcs
x=vcs.init()
f=cdms2.open(vcs.prefix+"/sample_data/clt.nc")
s=f("clt",time=slice(0,23))
gm = x.createboxfill()
levs = vcs.mkevenlevels(20,80)
gm.levels= levs
gm.fillareacolors = vcs.getcolors(levs)
d = x.plot(s,gm)
raw_input("Press enter")
x.animate.create(thread_it=False,min=20,max=80)
x.animate.fps(5000)
x.animate.run()
raw_input("Press enter to end")
x.animate.stop()
```

chaosphere2112:
I was testing out some of the tutorial scripts (and updating the animation one) when I ran into a couple of untested corner cases.

```python

import vcs, cdms2 as cdms, cdutil, time, os, sys

# Open data file:
filepath = os.path.join(vcs.prefix, 'sample_data/clt.nc')
cdmsfile = cdms.open( filepath )

# Extract a 3 dimensional data set and get a subset of the time dimension
data = cdmsfile['clt']

# Initialize VCS:
v = vcs.init()

# Plot data using the default isofill graphics method:
v.isofill( data )

# Initialize the animation
v.animate.create()
v.animate.run()
```

Which gives this error:
```
  File "/Users/fries2/uvcdat/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/Users/fries2/uvcdat/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/VTKAnimate.py", line 108, in run
    self.controller.draw_frame(allow_static = False, main_window_png = True)
  File "/Users/fries2/uvcdat/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/VTKAnimate.py", line 215, in draw_frame
    update_input(self)
  File "/Users/fries2/uvcdat/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/VTKAnimate.py", line 28, in update_input
    controller.vcs_self.backend.update_input(disp.backend,slab(*args),update=update)
  File "/Users/fries2/uvcdat/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/VTKPlots.py", line 1799, in update_input
    mapper.SetInputConnection(ports[i].GetOutputPort())
IndexError: list index out of range
```
It looks like the problem here is that isofill provides 2 vtk_backend_actors, but only 1 port in vtk_backend_countours.

The other issue is that the animation tutorial does this:

```python


import vcs, cdms2 as cdms, cdutil, time, os, sys

# Open data file:
filepath = os.path.join(vcs.prefix, 'sample_data/clt.nc')
cdmsfile = cdms.open( filepath )

# Extract a 3 dimensional data set and get a subset of the time dimension
data = cdmsfile['clt']

# Initialize VCS:
v = vcs.init()

# Plot data using the default isofill graphics method:
v.isofill( data )

# Initialize the animation
v.animate.create()
v.animate.run()
v.animate.stop()
v.animate.run()
v.animate.stop()
```

Which doesn't work at all; the Playback thread never gets restarted. I'm working on that issue right now; the isofill is probably more your realm.

aashish24:
@chaosphere2112 that would break VTk and OpenGL. Moving renderers like this from one context to another one is just very risky. I am not sure if you know much about OpenGL but there you have context specific to per window and when you do that, you are actually recreating the entire GL dataset and everything else again. Also, Since you are sharing the same actors with other window, now you are recreating it there as well. I would suggest, we fix the problem in some other way? 

Do you mind if we talk about it tomorrow morning? Let's do a hangout?

doutriaux1:
@aashish24 @chaosphere2112 let's not delay this over and over again. Can you meet today? @aashish24 we know we need to rewrite everything in a wotkflow way, it's scheduled for next release. But we need something for this one. Please meet with @chaosphere2112 today so we can move forward. @williams13 what do you think?

chaosphere2112:
@doutriaux1 I've just pushed the fix to one of the two problems I mentioned above (the second script I posted). I submitted a bug for the second one (#1141)

chaosphere2112:
@aashish24 I believe you when you say that this solution has risks–you're the expert, after all– but so far there haven't been any issues on most of the platforms we support (Mac OSX 10.8, Mac OSX 10.10, and Ubuntu). You get that I'm removing all of the renderers from the main window first, right? Everything is removed, added to a different window, and then a new renderer that just displays a PNG file is added to the main window.

chaosphere2112:
@aashish24 Also, I'm happy to talk with you whenever about this; I'd prefer sooner than later, as @doutriaux1 said, we're trying to get this done so we can actually have the features that we want in this release.

aashish24:
@doutriaux1 we have to very careful with this branch. This could potentially cause other issues. I will talk to @chaosphere2112 today.

aashish24:
@dlonie  can you talk to @chaosphere2112

dlonie:
**Re: Reusing an actor in multiple renderers**, see http://vtk.1045678.n5.nabble.com/Should-I-use-one-actor-per-renderer-renderwindow-or-can-they-share-the-same-actor-td4954889.html for a discussion of why this is a bad idea. It was possible a long time ago when VTK only utilized  the OpenGL fixed pipeline and display lists (though the latter had to be disabled), but this is no longer the case. Part of a large VTK modernization effort currently underway will completely break this as we move towards fully utilizing graphics hardware through VBOs/GLSL/etc. which are tied to a single GL context. It may works now, but it is not a supported usecase and will most definitely not work in the near future.

**Re: flickering:** If I understand the description correctly, the flickering is caused by `VTKPlots.png(...)` calling `hideGUI()` and `showGUI()` around the export to remove the widgets from the output. A better solution would be to either:
- add an option to `VTKPlots.png` to leave the GUI in. This wouldn't be the best, as the animation will keep the old GUI in the frames.
- Remove the GUI for the duration of the animation. If the stop/play/pause buttons are rendered in the GL widget, they should be moved into the uvcdat GUI for this approach. Alternatively, a progress dialog could pop up during the initial pass that generates the frames, and should have a cancel button in case the user wants to abort the png generation phase.
- If you must go the route of using a second offscreen renderer, don't just copy the actors onto it -- create a new vcs canvas and replay the plot commands onto it to reproduce the images.

chaosphere2112:
@dlonie 

1. We can't leave the GUI in the frames, because we export all of the PNGs to make the mp4 file when the user chooses to save, and if all of the GUI stuff stays in, nobody will ever use our mp4 export
2. We can't remove the GUI for the duration of the animation; when people use the code from the interactive python console, or a script that calls `interact()`, they don't have access to the QT GUI.
3. I'll give this method a shot and report back.

chaosphere2112:
@dlonie OK, I think the VCS replotting in a BG window works fine. It's not working right now in the UV-CDAT GUI, but that should be an issue I can solve before the end of the day.

chaosphere2112:
Ugh, running into weird vistrails behavior. Looks like I'll have to sort this all out tomorrow.

aashish24:
Also, seems like it has merge conflict. @chaosphere2112 I will let @dlonie review it tomorrow morning as well. Thanks for your patience. I appreaciate you addressing our concerns.

aashish24:
@dlonie can you review it?

chaosphere2112:
@aashish24 @dlonie It should be pretty much ready (besides the merge conflict). I think I found my issue with animating in the GUI, doing some testing.

chaosphere2112:
@aashish24 @dlonie On second thought, hold off for a bit; looks like there are some other issues with how I'm handling stuff in the GUI that needs to be cleaned up.

chaosphere2112:
@aashish24 @dlonie Ok, the issue I'm having is present universally; I'm just not handling window resize while animating right now. Easy enough to fix. Go ahead and review stuff, I'm just making a few tweaks to that.

Also, the issue I was having in the GUI was a VTK problem– the `QVTKRenderWindowInteractor` doesn't handle timers like VTK normally does. I've got a fix for that; should I submit a PR to uvcdat/vtk or vtk/vtk?

chaosphere2112:
@dlonie OK, give or take one small issue with how resizing is working (that I'm going to need to chat with @doutriaux1 about on Monday), this branch is ready for review.

chaosphere2112:
@aashish24 @dlonie Also, the fixes I mentioned for VTK are here: https://github.com/chaosphere2112/VTK/tree/qvtk_timer_fix

aashish24:
okay, thanks. We will get it merged today if all goes well.

chaosphere2112:
@aashish24 Please take a look at #1148; it looks like that's a VTK bug in handling offscreen rendering, which is the root of my current animation issues.

Packages/vcs/Lib/VTKAnimate.py
```diff
@@ -32,57 +83,189 @@ class VTKAnimationPlayback(animate_helper.AnimationPlayback):
...
+            karg.update(kargs)
+            if d.ratio is not None:
+                karg.update({"ratio":d.ratio})
+
+            canvas.plot(*parg, **karg)
+
+    def extract_renderers(self):
+        """
+        Pulls all non-background renderers from the main window
+        If background is true, it moves the renderers to the offscreen window in the creation thread.
```
dlonie: This docstring looks out of date

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1406,34 +1406,38 @@ def put_img_on_canvas(self,filename,zoom=1,xOffset=0,yOffset=0,*args,**kargs):
     cam.SetFocalPoint(xc+xoff,yc+yoff,0.)
     cam.SetPosition(xc+xoff,yc+yoff,d)
     ren.AddActor(a)
+    layer = self.renWin.GetNumberOfLayers() - 2
```
dlonie: Is it possible for layer to be less than 0 in this case? Might want to add a check if so.

Packages/vcs/Lib/VTKAnimate.py
chaosphere2112: I'll update this

Packages/vcs/Lib/VTKPlots.py
chaosphere2112: Good catch; I'll throw a max(0,...) check there.

chaosphere2112:
@dlonie Anything else?

dlonie:
Building/testing now :)

Is there a unit test that exercises this? If not, what's a good run through of this functionality that I can test?

chaosphere2112:
No unit test as of yet; best to test in three steps:

* By script
```python
import vcs, cdms2, sys, os
cfile = cdms2.open(os.path.join(sys.prefix, "sample_data", "clt.nc"))
x = vcs.init()
x.plot(cfile["clt"])
x.animate.create()
x.animate.run()
raw_input("Press enter to continue")
x.animate.stop()
name = raw_input("Enter name for animation: ")
x.animate.save(name)
```

* Via Interactive UI
    * Run lines 1:4 of above script
    * do `x.interact()`
    * Click "Open Configure"
    * Click "Open Animation"
    * Click "Run Animation"
    * Resize window
    * Click "Stop Animation"
    * Try "Step Forward" and "Step Backward"
    * Use the Time Slider
    * Change the Frames Per Second and hit "Run Animation"– if it's done a full loop of the images, it will play at whatever speed you specified (or slower, if it's still generating PNGs)
    * Save animation to a file, open in video player of choice (it's in .uvcdat/animation)
* Via UV-CDAT GUI
    * Plot the `clt` variable in the GUI, do basically the same things as you did in the interactive UI.

chaosphere2112:
@dlonie Let me know how the responsiveness is… and if you get some weird resizing artefacts (like the image not being quite the right aspect ratio), that's a known bug that I'm waiting on #1148 to fix.

dlonie:
Running the script gives me the following output, and the process ends before I see the prompts:

```
$ python test.py                                                                     
/ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py:3505: UserWarning: Unrecognized vcs plot keyword: cdmsfile, assuming backend (vtk) keyword                                                                                                            
  warnings.warn('Unrecognized vcs plot keyword: %s, assuming backend (%s) keyword'%(keyarg,self.backend.type))
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
X Error of failed request:  GLXBadDrawable
  Major opcode of failed request:  155 (GLX)
  Minor opcode of failed request:  11 (X_GLXSwapBuffers)
  Serial number of failed request:  112
  Current serial number in output stream:  114
```

chaosphere2112:
Huh. Can you step through it and see if it's the `plot`, `create`, or `run` call that's triggering the error? I'd assume the `create` or `run`.

dlonie:
It's in plot, the same happens when I replace everything after plot with interact.

chaosphere2112:
Huh. I'm going to go pester @doutriaux1 and see if he has any ideas.

dlonie:
BTW, do you know if you're using the X backend? Check the render window's classname (`renWin.GetClassname()`) -- the X backend will be named `vtkXOpenGLRenderWindow`.

If you see anything else, that may be why you're not seeing the error, at it looks to be X-specific.

chaosphere2112:
Nope, I'm on a mac. Definitely looks like I broke something about the plotting on x. I'll fix that– I've got a VM to test on.

doutriaux1:
@dlonie on my ubunut it is broken has well, so it does seem to be X11 related

chaosphere2112:
@dlonie @doutriaux1 Figured out the issue; a configureEvent was getting called, which led to a clear() call, which led to a premature render before the renWin was finished being configured. I made that not happen by initializing `_lastSize` in the `initialSize()` method, which works pretty well. Should be good for more testing now.

Also, @dlonie, for testing in the UV-CDAT GUI, you'll need a couple of unmerged PRs– UV-CDAT/VisTrails#23 and UV-CDAT/VTK#2

aashish24:
@chaosphere2112 I have few questions on UV-CDAT/VTK#2.

dlonie:
I'm getting a segfault during some OpenGL calls with this latest version (using the VTK/vistrails patches).

A call to `glMapBuffer` is returning NULL but not setting a gl error code, which is not supposed to happen...This makes me think it's a driver issue on my end. Can someone else test on linux? Otherwise we might need to scrap the threaded approach, as VTK isn't really intended to be used from multiple threads.

chaosphere2112:
@dlonie In which of the test cases are you having the issues? When in `interact()` mode, everything is executed using vtk Timers; I could possibly use those instead of threads in the command line mode, but I don't know if they would work if the `vtkRenderWindowInteractor` hadn't been `Start()`ed.

dlonie:
@chaosphere2112 This was using the testing script, without interact.

chaosphere2112:
OK, I'll give that a shot on my ubuntu VM and report back.

chaosphere2112:
Hm... the script seems to run fine on my ubuntu. @doutriaux1 mind giving the script above a shot?

doutriaux1:
@chaosphere2112 yep big time seg fault. But I didn't update my VTK do I need to?

chaosphere2112:
Shouldn't need to; mind if I come over and poke it with a stick?

dlonie:
If it's the same one I'm seeing, it's in `vtkPixelBufferObject`, and the returned value from `glMapBuffer` is `NULL`, but there is no opengl error set.

chaosphere2112:
The issue we were having was in cleaning up the offscreen render window; I've got a fix for that which I'm pushing shortly.

chaosphere2112:
@dlonie Can you try with the updated code?

dlonie:
Still seeing a segfault in the same spot here.

BTW, how is the threading implemented here? Is each frame being generated in a separate thread, or is each frame being generated successively in a single background thread?

chaosphere2112:
There are technically two background threads, one for creation and one for animation. 

On the command line, the creation thread is really just an object that handles the creation of new frames, and the animation thread requests the new frames. The performance for this can almost definitely be improved for command line users.

In `interact`, threads don't work reliably at all, so I don't use them. I set up a timer that will request a new frame every 1 / FPS milliseconds, and it will generate the new frames as that goes. I could, most likely, set up another timer to generate the frames, and then just fetch them as the animation timer needs them, but for the present I wanted to keep it simple.

dlonie:
The reason I ask is that my debugger showed 30 threads spawning and then closing almost immediately. This may have something to do with the `glMapBuffer` error, as the typical reason for it to fail is running out of contiguous space on the GPU, which could possibly happen if many simultaneous renderings were requested.

chaosphere2112:
Hm... that's pretty intense. At which point does it do this? During the animate.run()? Or during the animate.save()?

dlonie:
During plot :/

```
[Thread debugging using libthread_db enabled]                                                                                          
Using host libthread_db library "/usr/lib/libthread_db.so.1".                                                                          
plot                                                                                                                                   
/ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py:3492: UserWarning: Unrecognized vcs plot keyword: cdmsfile, assuming backend (vtk) keyword                                                                                                            
  warnings.warn('Unrecognized vcs plot keyword: %s, assuming backend (%s) keyword'%(keyarg,self.backend.type))
[New Thread 0x7fffc242c700 (LWP 25501)]                                                                                                
[Thread 0x7fffc242c700 (LWP 25501) exited]                                                                                             
[New Thread 0x7fffc1c2b700 (LWP 25502)]                                                                                                
[Thread 0x7fffc1c2b700 (LWP 25502) exited]                                                                                             
[New Thread 0x7fffc142a700 (LWP 25503)]                                                                                                
[Thread 0x7fffc142a700 (LWP 25503) exited]                                                                                             
[New Thread 0x7fffc0c29700 (LWP 25504)]                                                                                                
[Thread 0x7fffc0c29700 (LWP 25504) exited]                                                                                             
[New Thread 0x7fffc0428700 (LWP 25505)]                                                                                                
[Thread 0x7fffc0428700 (LWP 25505) exited]                                                                                             
[New Thread 0x7fffbfc27700 (LWP 25506)]                                                                                                
[Thread 0x7fffbfc27700 (LWP 25506) exited]                                                                                             
[New Thread 0x7fffbf426700 (LWP 25507)]                                                                                                
[Thread 0x7fffbf426700 (LWP 25507) exited]                                                                                             
[New Thread 0x7fffbf426700 (LWP 25508)]                                                                                                
[Thread 0x7fffbf426700 (LWP 25508) exited]                                                                                             
[New Thread 0x7fffbfc27700 (LWP 25509)]                                                                                                
[Thread 0x7fffbfc27700 (LWP 25509) exited]                                                                                             
[New Thread 0x7fffc0428700 (LWP 25510)]                                                                                                
[Thread 0x7fffc0428700 (LWP 25510) exited]                                                                                             
[New Thread 0x7fffc0c29700 (LWP 25511)]                                                                                                
[Thread 0x7fffc0c29700 (LWP 25511) exited]                                                                                             
[New Thread 0x7fffc242c700 (LWP 25512)]                                                                                                
[Thread 0x7fffc242c700 (LWP 25512) exited]                                                                                             
[New Thread 0x7fffc1c2b700 (LWP 25513)]                                                                                                
[Thread 0x7fffc1c2b700 (LWP 25513) exited]                                                                                             
[New Thread 0x7fffc142a700 (LWP 25514)]                                                                                                
[Thread 0x7fffc142a700 (LWP 25514) exited]                                                                                             
[New Thread 0x7fffc142a700 (LWP 25515)]                                                                                                
[Thread 0x7fffc142a700 (LWP 25515) exited]                                                                                             
[New Thread 0x7fffc1c2b700 (LWP 25516)]                                                                                                
[Thread 0x7fffc1c2b700 (LWP 25516) exited]                                                                                             
[New Thread 0x7fffc242c700 (LWP 25517)]                                                                                                
[Thread 0x7fffc242c700 (LWP 25517) exited]                                                                                             
[New Thread 0x7fffc0c29700 (LWP 25518)]                                                                                                
[Thread 0x7fffc0c29700 (LWP 25518) exited]                                                                                             
[New Thread 0x7fffc0428700 (LWP 25519)]                                                                                                
[Thread 0x7fffc0428700 (LWP 25519) exited]                                                                                             
[New Thread 0x7fffbfc27700 (LWP 25520)]                                                                                                
[Thread 0x7fffbfc27700 (LWP 25520) exited]                                                                                             
[New Thread 0x7fffbf426700 (LWP 25521)]                                                                                                
[Thread 0x7fffbf426700 (LWP 25521) exited]                                                                                             
[New Thread 0x7fffbf426700 (LWP 25522)]
[Thread 0x7fffbf426700 (LWP 25522) exited]
[New Thread 0x7fffbfc27700 (LWP 25523)]
[Thread 0x7fffbfc27700 (LWP 25523) exited]
[New Thread 0x7fffc0428700 (LWP 25524)]
[Thread 0x7fffc0428700 (LWP 25524) exited]
[New Thread 0x7fffc0c29700 (LWP 25525)]
[Thread 0x7fffc0c29700 (LWP 25525) exited]
[New Thread 0x7fffc242c700 (LWP 25526)]
[Thread 0x7fffc242c700 (LWP 25526) exited]
[New Thread 0x7fffc1c2b700 (LWP 25527)]
[Thread 0x7fffc1c2b700 (LWP 25527) exited]
[New Thread 0x7fffc142a700 (LWP 25528)]
[Thread 0x7fffc142a700 (LWP 25528) exited]
create
[New Thread 0x7fffc142a700 (LWP 25529)]
run
[Thread 0x7fffc142a700 (LWP 25529) exited]
[New Thread 0x7fffc142a700 (LWP 25530)]
Press enter to continue
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc142a700 (LWP 25530)]
0x00007fffcb3bb9b3 in vtkUpload3D<unsigned char>::Upload (pboPtr=0x0, inData=0x1dd9640 "", dims=0x7fffc14286b0, numComponents=4, 
    continuousIncrements=0x7fffc1428690, components=0, componentList=0x0)
    at /ssd/src/uvcdat-build/build/VTK/Rendering/OpenGL/vtkPixelBufferObject.cxx:351
```

chaosphere2112:
Weird. In theory it shouldn't be touching the animation code until `x.animate.create()` gets called. Can you throw a `pdb` breakpoint into `vcs/Lib/animate_helper.py`'s `AnimationController.create` function and see if that's happening before the segfault?

dlonie:
It segfaults before it gets to that method.

chaosphere2112:
OK, so it's not the animation code, it'll be the changes I've made to `VTKPlots.py`'s configureEvent, most likely. In essence, the same issue that I thought I had fixed yesterday. Hm... can you do a pdb breakpoint in that function, and step through with 'n' till you find the line that causes the segfault?

chaosphere2112:
@dlonie I'd recommend sticking the breakpoint after the `_lastSize == size` check

dlonie:
Here's the stack for the render call that triggers the segfault:

```
-> self.__bootstrap_inner()
  /ssd/src/uvcdat-build/install/lib/python2.7/threading.py(810)__bootstrap_inner()
-> self.run()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(713)run()
-> self.controller.draw_frame()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(79)draw_frame()
-> self.vcs_self.backend.update_input(disp.backend,slab(*args),update=True)
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(1863)update_input()
-> self.renWin.Render()
```

chaosphere2112:
Oh, funny. OK... so I guess the run thread is somehow getting triggered prematurely. Can you stick a breakpoint in `AnimationPlayback.__init__` in animate_helper.py and give me the backtrace?

dlonie:
Actually, I'll back up a bit -- I'd put the breakpoint in the wrong `create()` method. The controller's create method is getting called before the segfault:

During Canvas.clear (during `plot` in the script):

```
  /home/dl/prm/UVCDAT/1113-interact-animate/test.py(5)<module>()
-> x.plot(cfile["clt"])
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py(2599)plot()
-> a = self.__plot( arglist, keyargs )
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py(3663)__plot()
-> returned_kargs = self.backend.plot(*arglist,**keyargs)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(392)plot()
-> self.createRenWin(**kargs)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(295)createRenWin()
-> self.createDefaultInteractor(self.renderer)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(274)createDefaultInteractor()
-> self.vcsInteractorStyle.SetCurrentRenderer( ren )
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(225)configureEvent()
-> self.canvas.clear()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py(3826)clear()
-> if self.animate.created():
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(751)created()
-> return self.animation_created
```

Again in configure event (still in `plot`):

```
  /home/dl/prm/UVCDAT/1113-interact-animate/test.py(5)<module>()                                                                       
-> x.plot(cfile["clt"])                                                                                                                
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py(2599)plot()                                                  
-> a = self.__plot( arglist, keyargs )                                                                                                 
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/Canvas.py(3663)__plot()
-> returned_kargs = self.backend.plot(*arglist,**keyargs)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(392)plot()
-> self.createRenWin(**kargs)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(295)createRenWin()
-> self.createDefaultInteractor(self.renderer)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(274)createDefaultInteractor()
-> self.vcsInteractorStyle.SetCurrentRenderer( ren )
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(229)configureEvent()
-> if self.canvas.animate.created():
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(751)created()
-> return self.animation_created
```

Again during playback (now during `run` in the script):

```
  /home/dl/prm/UVCDAT/1113-interact-animate/test.py(5)<module>()
-> x.plot(cfile["clt"])
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(782)playback()
-> if (self.created() and
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(751)created()
-> return self.animation_created
```

Then the `__init__` method on playback is called during `run`:

```
  /home/dl/prm/UVCDAT/1113-interact-animate/test.py(5)<module>()
-> x.plot(cfile["clt"])
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(784)playback()
-> self.playback_thread = self.AnimationPlayback(self)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(33)__init__()
-> animate_helper.AnimationPlayback.__init__(self,controller)
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(703)__init__()->None
-> import pdb; pdb.set_trace();
```

Then it segfaults in the `self.renWin.Render()` call posted earlier.

chaosphere2112:
That other function you are referring to– `created`– is just a check to see if the animation has already been initialized; it doesn't initialize anything itself, just checks a boolean flag.

I am a little confused by that last stacktrace, though. Can you remove everything after the plot in the test script? It seems super weird that the stacktrace skips from x.plot() to the middle of `animate.playback` (which `animate.run` is aliased to).

dlonie:
Ah good catch, I still had it in the wrong spot >.<

In any case, I can't reproduce that last stack trace now. Now I get:

```
  /home/dl/prm/UVCDAT/1113-interact-animate/test.py(10)<module>()
-> x.animate.run()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(783)playback()
-> self.playback_thread = self.AnimationPlayback(self)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(33)__init__()
-> animate_helper.AnimationPlayback.__init__(self,controller)
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(703)__init__()->None
-> import pdb; pdb.set_trace();
```

I have a hard time believing that the pdb got confused, but I'm not sure what else would explain that....

chaosphere2112:
OK, so the animation is getting created and running. That makes more sense! Whew.

Looking at the stack trace of the render call that fails...

```
-> self.__bootstrap_inner()
  /ssd/src/uvcdat-build/install/lib/python2.7/threading.py(810)__bootstrap_inner()
-> self.run()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/animate_helper.py(713)run()
-> self.controller.draw_frame()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(79)draw_frame()
-> self.vcs_self.backend.update_input(disp.backend,slab(*args),update=True)
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(1863)update_input()
-> self.renWin.Render()
```
This looks a little weird to me. The line in VTKAnimate that triggers this render has changed in this branch. It looks like you're using the [master](https://github.com/UV-CDAT/uvcdat/blob/14f9225b92b7b01d378fc4594112c3893527d2a5/Packages/vcs/Lib/VTKAnimate.py#L79) branch rather than [this one](https://github.com/chaosphere2112/uvcdat/blob/d894672568cef2bbfd2f943158d7e21b2a21c343/Packages/vcs/Lib/VTKAnimate.py#L79); is that possible?

dlonie:
D'oh, you're right -- I had switched to Charles' tick label branch and forgot to switch back.

It still segfaults in the same spot, though. Here's the correct trace:

``` 
 /ssd/src/uvcdat-build/install/lib/python2.7/threading.py(783)__bootstrap()
-> self.__bootstrap_inner()
  /ssd/src/uvcdat-build/install/lib/python2.7/threading.py(810)__bootstrap_inner()
-> self.run()
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(93)run()
-> self.controller.draw_frame(allow_static = True, render_offscreen = False, main_window_png = True)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(246)draw_frame()
-> update_input(self.vcs_self, self._number_of_dims_used_for_plot, frame_num)
  /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKAnimate.py(29)update_input()
-> canvas.backend.update_input(disp.backend, slab(*args), update=update)
> /ssd/src/uvcdat-build/install/lib/python2.7/site-packages/vcs/VTKPlots.py(1870)update_input()
-> self.renWin.Render()
```

chaosphere2112:
@dlonie I removed some code which I think was unnecessary (updating all of the actors by hand in `update_info`). I don't know that this will fix your segfault, but, hey, free processor cycles. Maybe you could try enabling debug on the `renWin` when you hit that breakpoint? Maybe that will give us some insight.

chaosphere2112:
@dlonie Out of curiosity, what distro are you using? I'll try and get a VM up and running to see if I can reproduce.

dlonie:
@chaosphere2112 I'm on Arch Linux. I'll try out the latest changes and get back to you.

dlonie:
Still segfaults on the latest.

BTW, I wouldn't spend too much time trying to reproduce this in a VM. The GL implementation is acting buggy, and I don't believe it's possible to use the nvidia drivers from a VM.

chaosphere2112:
Looks like our firewall doesn't like talking to Arch's site, so I can't even get my hands on the ISO to set up a VM.

dlonie:
Does anyone else have a nVidia card with up-to-date linux drivers to test this? That'd be the best shot for reproducing this.

If it's just my machine failing, I won't hold up the merge, but I still have reservations about running VTK in multiple threads.I know I keep harping on this, but this just isn't something we officially support or test for in VTK. If it starts causing problems down the road we may end up having to revisit this issue again.

chaosphere2112:
No worries about harping on this; I know running code that wasn't designed to handle concurrency in a concurrent environment is a bad idea.

That said, this isn't concurrent. The only situation when threads are actually being used for their "real" purpose is between `x.animate.run()` and `x.animate.stop()`. The creation thread's run function is empty– as I said earlier, it's just an object that manages the creation and retrieval of frames, as prompted by other objects. We could set it up so any calls to the canvas while the animation is running stops the animation, if that would help your concerns– that's just about the only situation I can picture where more than one thread would be accessing the render window at once.

chaosphere2112:
@dlonie In fact, I'm going to go ahead and do that anyway; right now that's just a weird undefined behavior.

chaosphere2112:
@dlonie OK, now it should be able to plot while animating (will stop the animation). Are there any other guards against multithreaded VTK that you have concerns about?

dlonie:
@chaosphere2112 Not off the top of my head.

dlonie:
@chaosphere2112 @doutriaux1 @aashish24 Looks like we'll need someone else to merge this, it doesn't work on my machine so I can't test it.

I get segfaults resulting from a poorly behaved gl call, so it looks like a driver issue on my end, not necessarily an issue with the code.

jbeezley:
I can give it a shot.

chaosphere2112:
@jbeezley Great, you can check out the instructions I gave– script for what to do is [here](https://github.com/UV-CDAT/uvcdat/pull/1113#issuecomment-82431373), patches you need are [here](https://github.com/UV-CDAT/uvcdat/pull/1113#issuecomment-82620707)

chaosphere2112:
My most recent commit fixes an issue where button labels would occasionally get misaligned; it only happens on this branch, so it was result of some change in here.

jbeezley:
It works for me.  I have a couple of comments.

1. The output movies rendered one captured image per frame at 25 fps, which was much faster than realtime.  It would be nice to duplicate the frames as necessary or reduce the output fps.
2.  The record button dimming doesn't make it very clear that the recording is taking place or how to stop it.  I would be better change the text of "record" to something like "stop recording"

chaosphere2112:
@jbeezley 

1. The output FPS should be controllable by the FPS slider, which is, admittedly, non-intuitive.
2. Currently the saving process is not stoppable, though that would be a good thing to do. I can add that, I think. I'll give it a shot today and see what happens.

jbeezley:
I guess I don't know exactly what was going on then.  I though it was recording my interactions until I hit the record button again.  I also thought the FPS slider was just to change the on screen animation.  It didn't seem to affect the video file.  I can check again to make sure.

chaosphere2112:
What are you plotting? That may be how the 3D stuff works, not the 2D.

doutriaux1:
@chaosphere2112 did we double check that the fps slider is also used by the save function?

chaosphere2112:
@doutriaux1 Yup; [here](https://github.com/chaosphere2112/uvcdat/blob/e5b413c36507dc4bebfffedfd7767291ee83327c/Packages/vcs/Lib/configurator.py#L576-L582) is where the `animation_speed` variable is managed by the FPS slider, 
 [here](https://github.com/chaosphere2112/uvcdat/blob/e5b413c36507dc4bebfffedfd7767291ee83327c/Packages/vcs/Lib/configurator.py#L552-L574) is where the FPS gets set with the `animation_speed` variable, and [here](https://github.com/chaosphere2112/uvcdat/blob/e5b413c36507dc4bebfffedfd7767291ee83327c/Packages/vcs/Lib/animate_helper.py#L1024-L1025) is where the FPS variable is retrieved while saving. Also, if you change the animation speed, you get videos of different length– I tested making movies at 24 fps and 3 fps from the command line, and 30 and 1 fps from the UI, and they all generated appropriate length films. @jbeezley, do you mean that the default speed is too high? It looks like it's set to 10 in [here](https://github.com/chaosphere2112/uvcdat/blob/e5b413c36507dc4bebfffedfd7767291ee83327c/Packages/vcs/Lib/animate_helper.py#L658-L663).

With regards to saving out the animation, it does say "Save Animation", not "Record Animation". Though, again, I agree that the UI is less than optimal. The intended effect is to save out the entire time duration of the plot to an MP4 file, animated at whatever the current FPS is.

jbeezley:
I was using the 3D plot.  With a 2D plot, I don't see a way to using the interaction in the uvcdat gui.  Calling `x.interact()` from the CLI.  I can do everything you mentioned, but if I click save animation a second time, the animation loops infinitely and I have to kill the process.

doutriaux1:
@chaosphere2112 on my mac 10.9 command line and interact work fine.
GUI though doesn't quite work. First of all it instantly bring up the widgets, did we not decide to do it only after we click the "pencil" buttons?
Then when I run the animation the ticks are all weirded out.

I'm using this branch for vcs, vtk: uvcdat-master with your timer fix, vistrails: your branch on pr

Am i missing sometihng?

chaosphere2112:
@jbeezley Sorry, I ran into that bug as well. I have a fix for it, I'll push that up momentarily. To get at the controls in the GUI, you have to click the Configure pencil in the toolbar above the plot.

@doutriaux1 It's working on my machine... Are you sure you've got the latest of all of those?

chaosphere2112:
@jbeezley OK, you can cancel a save now, and should be able to save multiple times without breaking things.

@doutriaux1 You still having issues? Are you in at all today?

jbeezley:
Okay, I understand now.  Looks to be working for me.

doutriaux1:
@chaosphere2112 yes I'm sure I got latest of all. But will do a build from scratch just to be sure. And yes I'm in but turned off email so I can get things done.

chaosphere2112:
Huh, wacky. If you want to debug it a bit, you can stick a breakpoint in `canvas.configure`, and see what's calling it.

doutriaux1:
@chaosphere2112 interesting. When I click the pen and close the configure panel that makes the widgets disappear. Then clicking on the pen makes them reappear.

chaosphere2112:
So you've definitely got the right code in there, if it's happening like that. Are you just plotting clt?

doutriaux1:
yep. Wonder why it pops up with first plot.

chaosphere2112:
Did you get anything with the pdb / set_trace() in the canvas.configure method?

doutriaux1:
@chaosphere2112 does this helps:
```
 ~~~~~~~~~~~~~~~~~~ create plot functions:  [('graphicsMethodName', ['a_boxfill']), ('plotOrder', ['1']), ('boxfill_type', ['linear']), ('color_1', [16]), ('color_2', [239]), ('datawc_calendar', [135441]), ('datawc_timeunits', ['days since 2000']), ('datawc_x1', [1e+20]), ('datawc_x2', [1e+20]), ('datawc_y1', [1e+20]), ('datawc_y2', [1e+20]), ('levels', [[1e+20, 1e+20]]), ('ext_1', ['n']), ('ext_2', [False]), ('fillareacolors', [None]), ('fillareaindices', [None]), ('fillareastyle', ['solid']), ('legend', [None]), ('level_1', [1e+20]), ('level_2', [1e+20]), ('missing', [1]), ('projection', ['linear']), ('xaxisconvert', ['linear']), ('xmtics1', ['']), ('xmtics2', ['']), ('xticlabels1', ['*']), ('xticlabels2', ['*']), ('yaxisconvert', ['linear']), ('ymtics1', ['']), ('ymtics2', ['']), ('yticlabels1', ['*']), ('yticlabels2', ['*'])]
/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.py:3512: UserWarning: Unrecognized vcs plot keyword: cdmsfile, assuming backend (vtk) keyword
  warnings.warn('Unrecognized vcs plot keyword: %s, assuming backend (%s) keyword'%(keyarg,self.backend.type))
/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.py:3512: UserWarning: Unrecognized vcs plot keyword: cell_coordinates, assuming backend (vtk) keyword
  warnings.warn('Unrecognized vcs plot keyword: %s, assuming backend (%s) keyword'%(keyarg,self.backend.type))
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/uvcdat.py", line 131, in <module>
    v = app.exec_()
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/packages/spreadsheet/spreadsheet_tab.py", line 961, in dropEvent
    sheetName, row, col))
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/packages/spreadsheet/spreadsheet_tabcontroller.py", line 1064, in plotDropped
    self.emit(QtCore.SIGNAL("dropped_plot"), info)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/project_controller.py", line 776, in plot_was_dropped
    self.check_update_cell(sheetName,row,col, update)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/project_controller.py", line 910, in check_update_cell
    self.update_cell(sheetName, row, col, reuse_workflow)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/project_controller.py", line 992, in update_cell
    reuse_workflow)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/project_controller.py", line 1039, in update_workflow
    self.update_plot_configure(sheetName, row, column)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/project_controller.py", line 847, in update_plot_configure
    plot_prop.set_controller(self)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/plot.py", line 30, in set_controller
    self.updateProperties(None)
  File "/lgm/uvcdat/2015-03-20/vistrails/vistrails/gui/uvcdat/plot.py", line 58, in updateProperties
    canvas.configure()
  File "/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.py", line 981, in configure
    traceback.print_stack()
```

chaosphere2112:
@doutriaux1 Yeah; can you do a git pull in vistrails? I forgot to push a commit.

doutriaux1:
@chaosphere2112 ok widgets now work, but I still get a weird ticks behaviour when runnig anim see attached png

ticks_off

doutriaux1:
oh and closing edit then clicking on pen again and clicking run anim again leads to this:
```
  File "/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/VTKAnimate.py", line 139, in modified
    shutil.rmtree(os.path.dirname(self.animation_files[0]))
  File "/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 239, in rmtree
    onerror(os.listdir, path, sys.exc_info())
  File "/lgm/uvcdat/2015-03-20/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 237, in rmtree
    names = os.listdir(path)
OSError: [Errno 2] No such file or directory: '/Users/doutriaux1/.uvcdat/bbd027a9bafe410f3900b5f1a700c7a780d1d03a'
```

chaosphere2112:
Hmm... I'm not having exactly those issues, but definitely similar ones. I think there are some resizing quirks to work out on Monday.

chaosphere2112:
@doutriaux1 Latest commit should fix issue where animation starts off with ticks in wrong places, and will speed up resizing the window while animating; since resizing the background render window doesn't work well, I'm just rendering pngs at double the original resolution, and letting them get scaled appropriately for now. When we get #1148 fixed, I'll update it to work more like it ought to. I'm still working on the exception after closing the animation.

chaosphere2112:
@doutriaux1 Fixed the exception, helps to actually put the plot renderers back on the screen when configuration ends.

doutriaux1:
@chaosphere2112 ticks are fixed but resize doesn't resize while animating any longer.

doutriaux1:
@chaosphere2112 of course works, user error, sorry about the noise. @dlonie @jbeezley @aashish24  it's ready to go in as soon as we have your blessing.

dlonie:
Still segfaults for me, but @aashish24 said he'll take a look on his Ubuntu.

aashish24:
I am thinking now that we can merge it and see if we can reproduce the crash on other systems. I will try it tomorrow.

aashish24:
Lets wait if @williams13 has any suggestions.

chaosphere2112:
@doutriaux1 mentioned yesterday that he was just about ready to merge this so we can get as many people testing it as possible to work out the last few kinks.

doutriaux1:
@chaosphere2112 merging now.

PR #1115, opened 2015-03-05
adding ipython created cyclical deps

doutriaux1:


durack1:
@doutriaux1 I'm not sure we want spyder without ipython (in fact ipython is likely a hard dependency) - is there a way to get this cyclical dependency thing sorted out in 2.3+ - Should an issue be created for this?

doutriaux1:
No need to. This will be solved in 2.3 when getting rid of pip

durack1:
@doutriaux1 ok great - is there a pip issue for 2.3 which this should be linked to?

PR #1119, opened 2015-03-05
Isofill on pointdata struct grid

doutriaux1:
@aashish24  @dlonie here it is
goes with: https://github.com/UV-CDAT/uvcdat-testdata/pull/23


dlonie:
LGTM

PR #1129, opened 2015-03-07
Issue1128 durack1 cython0.22 update

durack1:
@doutriaux1 @aashish24 there's no rush to merge this - 2.3 would be fine..

PR #1131, opened 2015-03-07
Issue1126 durack1 matplotlib1.4.3 update

durack1:
@doutriaux1 @aashish24 there's no rush to merge this - 2.3 would be fine.. I've also removed the matplotlib_$releaseCount_min text from this file, not certain if this was used outside somewhere?


durack1:
Seems like there are some cmake changes which may cause problem here?
```
CMake Warning (dev) at CMake/cmake_modules/ExternalProject.cmake:199 (if):
Policy CMP0054 is not set: Only interpret if() arguments as variables or
keywords when unquoted. Run "cmake --help-policy CMP0054" for policy
details. Use the cmake_policy command to set the policy and suppress this
warning.
Quoted keywords like "COMMAND" will no longer be interpreted as keywords
when the policy is set to NEW. Since the policy is not set the OLD
behavior will be used.
Call Stack (most recent call first):
CMake/cmake_modules/ExternalProject.cmake:1555 (_ep_parse_arguments)
CMake/cdat_modules/pkgconfig_external.cmake:6 (ExternalProject_Add)
CMakeLists.txt:636 (include)
This warning is for project developers. Use -Wno-dev to suppress it.
[INFO] Package --- readline/READLINE/6.2 --- will be built
[INFO] Package --- Python/PYTHON/2.7.9 --- will be built
[INFO] Package --- NUMPY/NUMPY/1.9.0 --- will be built
[INFO] Package --- setuptools/SETUPTOOLS/1.4 --- will be built
[INFO] Package --- pip/PIP/1.2.1 --- will be built
CMake Warning (dev) at CMake/cdat_modules/pip_external.cmake:8 (include):
Policy CMP0053 is not set: Simplify variable reference and escape sequence
evaluation. Run "cmake --help-policy CMP0053" for policy details. Use the
cmake_policy command to set the policy and suppress this warning.
For input:
'@cdat_CMAKE_BINARY_DIR@/cdat_common_environment.cmake'
the old evaluation rules produce:
'/Users/travis/build/UV-CDAT/_build/CMake/cdat_common_environment.cmake'
but the new evaluation rules produce:
'@cdat_CMAKE_BINARY_DIR@/cdat_common_environment.cmake'
Using the old result for compatibility since the policy is not set.
Call Stack (most recent call first):
CMakeLists.txt:636 (include)
This warning is for project developers. Use -Wno-dev to suppress it.
```

durack1:
It seems that the matplotlib_$releaseCount_min text is used in `CMake/cmake_modules/FindMatplotlib.cmake` - not sure this is useful anymore?

durack1:
@aashish24 @doutriaux1 the only thing that has changed here is a tarball/md5 update - which is not even built as part of the travis tests.

Why is the build failing then?

durack1:
This is mostly the same issue with #1129 - travis failure rather than problems with the PR

aashish24:
@durack1 don't worry about it. We are revamping the build system. Lets try to get this one in post 2.2

durack1:
@aashish24 @doutriaux1 if it was possible it'd be great to get this patch update into 2.2.. But if you guys are trying to simplify things 2.3 would be ok I suppose..

Issue #1093, opened 2015-03-02
Editing a label, rotate to ~311 degrees, open font menu -> segfault


chaosphere2112:
When you edit a text object and rotate it to somewhere around 311 degrees and open the font menu, we segfault.

Discussion on #1081

doutriaux1:
@dlonie it might be VTK issue so assigning to you.

chaosphere2112:
@dlonie Any update on this? I'm probably going to be doing a workaround for this issue tomorrow if nothing looks promising.

dlonie:
@chaosphere2112 I haven't had a chance to look into it yet. Feel free to do a workaround.

aashish24:
Moving this to 2.3 for now as we have the workaround for 2.2

chaosphere2112:
:+1:

dlonie:
Is this still an issue? I'm trying to reproduce, but can you clarify what you mean by "font menu"? I don't see this. So far I've:

1) Plotted clt on default boxfill
2) Clicked the pencil to start editing
3) Select text
4) Show angle
5) Adjust to 311 degrees

Not sure what's next.

chaosphere2112:
@dlonie I removed the font menu (which was a button you could click in the configure menu to open a list of all of the fonts available, styled using those fonts) and replaced with one button you click through to cycle which font you're using. It's not ideal, but it'll work for now. We can look into this further in a week or two.

dlonie:
Aha, that would explain it :dancer: 

Let me know when this is ready to look into again.

dlonie:
Actually, I just triggered a segfault by clicking through the available font faces. I'll see if this is the same one (need to recompile so python's unicode matches my debugger...)

chaosphere2112:
Ok, cool; let me know if you find anything.

dlonie:
It is the same crash. Interesting, because the docs say face->size is always initialized for new fonts (and indeed, the freetype sources do this). I'll poke around a bit and see what I can find.

dlonie:
Found the issue:

The FreeType library sets `face->size` to an appropriate FT_Size object when it loads the font face. However, once it is loaded into FreeType's internal caching system, it deletes this object and sets `face->size` to 0. This modified version is what FreeType gives to VTK when we request a face from the cache manager.

Then we query the kerning information, and an unguarded dereference of `face->size` occurs, triggering the segfault.

[...]

PR #1135, opened 2015-03-09
Workaround for #1093

chaosphere2112:
Eliminates the scenario that leads to the segfault in #1093 by replacing the font menu with a font button. Instead of getting previews of all of the fonts, you get a single button that cycles through the different fonts. Less good, but less crash-prone.

I extracted font selection out into a separate class so when we actually fix the root of #1093 it should be pretty easy to bring back the more complicated but more user-friendly UI. We can either leave #1093 open (and move to 2.3) or closed (and create an enhancement request to add a menu for fonts); I'd lean towards the latter.

### DIFF
--- a/Packages/vcs/Lib/editors/font.py
+++ b/Packages/vcs/Lib/editors/font.py
@@ -1,0 +1,16 @@
+import vcs
+
+class FontEditor(object):
+    def __init__(self, toolbar, font_setter, current_font="default"):
+
+        self.toolbar = toolbar
+        self.set_font = font_setter
+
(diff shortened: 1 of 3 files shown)


chaosphere2112:
@doutriaux1 @aashish24

doutriaux1:
hum... on ubunutu after resizing it gets all weirded out and clicking show angle doesn't do anything
screenshot from 2015-03-09 13 35 16

chaosphere2112:
@doutriaux1 The resizing issues should be addressed in #1113; do you want to hold off on merging this till then, or can we put it in now? The issues weren't really related to this PR.

Issue #1126, opened 2015-03-07
Update matplotlib to 1.4.3


durack1:
We are a patch release behind matplotlib latest (1.4.2 vs 1.4.3).

@doutriaux1 @aashish24 want to assign me to this task as a 2.3 milestone?

PR #1137, opened 2015-03-10
tweaks for travis

doutriaux1:
@aashish24 @dlonie just testing please don't even bother looking at it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: c++
 before_install:
     - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo apt-get install -y gfortran; fi
     - if [ "$TRAVIS_OS_NAME" = "linux" ]; then pushd $HOME && mkdir cmake3.1 && cd cmake3.1 && (curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-x86_64.tar.gz" | gunzip -c | tar x) && cd cmake-*/bin && export PATH="${PWD}:${PATH}"; popd; fi
-    - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update ; brew outdated cmake || brew upgrade cmake ; fi
+    - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update ; brew outdated cmake || brew upgrade cmake ; brew install expat ; fi
     - cmake --version
 
 script:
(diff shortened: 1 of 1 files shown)


durack1:
@doutriaux1 another one of these weird timeouts again [LINK](https://travis-ci.org/UV-CDAT/uvcdat/builds/53841592).. Who knows if there was a problem or not..

durack1:
@doutriaux1 this should probably go into the default `.travis.yml` file:

http://docs.travis-ci.com/user/build-timeouts/#Build-times-out-because-no-output-was-received

.travis.yml
```diff
@@ -16,6 +16,6 @@ script:
     - cd ..
     - mkdir _build
     - cd _build
-    - cmake -DGIT_PROTOCOL=git:// -DCDAT_BUILD_MODE=LEAN -DCDAT_BUILD_GRAPHICS=ON -DCDAT_BUILD_SCIPY=OFF ../uvcdat
+    - cmake -DGIT_PROTOCOL=git:// -DCDAT_BUILD_MODE=DEFAULT -DCDAT_BUILD_GUI=OFF -DCDAT_BUILD_SCIPY=OFF ../uvcdat
```
aashish24: Given the time on travis, we can only build in LEAD mode or it will fail most of the time.

durack1:
@aashish24 I don't think this is true, with travis ci we have 120 mins.. It also seems that the problem is more related to timeouts within the install (no activity for 20 mins) rather than exceeding the 120 min limit

aashish24:
@durack1  please see below (http://docs.travis-ci.com/user/build-configuration). In any case, I want to keep the build process shorted so that we can provide quick feedback. Like I said, I am working on another solution until then, please don't change the travis-ci. 

----

Because it is very common to see test suites or before scripts to hang up, Travis CI has hard time limits. If a script or test suite takes longer to run, the job will be forcefully terminated and you will see a message about this in your build log.

With our current timeouts, a job will be terminated if it's still running after 50 minutes (120 minutes on travis-ci.com), or if there hasn't been any log output in 10 minutes.

Some common reasons why test suites may hang up:

Waiting for keyboard input or other kind of human interaction
Concurrency issues (deadlocks, livelocks and so on)
Installation of native extensions that take very long time to compile
There is no such timeout for a build; a build will run as long as all the jobs do.

durack1:
@aashish24 the issue was opened as it appeared the travis config (rather than any other change) was causing the build failures.. It seems fairly useless to have tests that are run and are failing if the only part failing is travis itself, rather than due to changes in the PR request being tested.

In #1126 in particular this is not even built under `LEAN` conditions, so it just shouldn't fail

doutriaux1:
@aashish24 the first comment says to you to not worry about this one. This is just @durack1 trying out things, when he's ready we will ask you to review don't worry. Re-opening so that the changes he makes get tested on travis.

aashish24:
things are tested on travis whether on not it has a pull request. Travis test every commit. Closing it for that reason.

PR #1144, opened 2015-03-12
Issue 587 projected ticks

doutriaux1:
@williams13 @aashish24 @chaosphere2112 I think this should go in for the rlease. It's better than what's in master.

BUT: polar projection have ticks drawn behind data. @dlonie can't look at it until later, but it souldn't be a show stopper.

Also I didn't run ctst, but for sure ALL projection tests will fail because a new baseline is needed for projected plots. @chaosphere2112 or @aashish24 can generate the new baseline.

PR #1154, opened 2015-03-18
Fix missing displays for animation/interaction

chaosphere2112:
#1144 added a slight regression– it removed all of the text actors from the display plot's backend. This fixes that problem


doutriaux1:
passes ctest for me as well.

PR #1161, opened 2015-03-25
VCS2D Animation Speedup

chaosphere2112:
Big speedup of animation in 2D– just had to stop saving PNGs while running the animation, and removed some redundant `Render` calls.

@doutriaux1 @williams13 @aashish24

### DIFF
--- a/Packages/vcs/Lib/VTKAnimate.py
+++ b/Packages/vcs/Lib/VTKAnimate.py
@@ -100,7 +100,7 @@ class VTKAnimationPlayback(animate_helper.AnimationPlayback):
       while not self.is_stopped():
           self.wait_if_paused()
 
-          self.controller.draw_frame(allow_static = True, render_offscreen = False, main_window_png = True)
+          self.controller.draw_frame(allow_static = False, render_offscreen = False, main_window_png = False)
 
           self.controller.frame_num += 1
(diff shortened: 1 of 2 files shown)


doutriaux1:
@chaosphere2112 that works great! Thank you! The one thing though is that it kills the zoom feature.
@williams13 do we want to keep it? @williams13 the other option is to turn on pngs (slow) when zooming? Your thoughts.

williams13:
Let's not do the zooming for this release. We will revisit for the next release.

aashish24:
@chaosphere2112 when does it write the PNGs then?

chaosphere2112:
@aashish24 It writes the PNGs when you call `canvas.animate.save("path.mp4")`; takes longer to save out the animation, but we get >30FPS when running, so it seems like a useful tradeoff; @williams13 and @doutriaux1 indicated that the most common use case is to just run the animation to take a look at it, and then to manually generate PNGs in which you manipulate a variety of things, then call `canvas.ffmpeg` on those PNGs.

aashish24:
okay sounds good. Earlier we had a diffirirent strategy as we suggested not to write file while we run the animation. I totally support this approach.

doutriaux1:
@williams13 go ahead and rebuild

PR #1177, opened 2015-03-26
Fixed some VTK error messages that showed up when ending configure

chaosphere2112:
Should hopefully fix @williams13's VTK error messages; managed to reproduce them locally and eliminated them. They cropped up when closing the configure menu when a bunch of plots were open.

### DIFF
--- a/Packages/vcs/Lib/configurator.py
+++ b/Packages/vcs/Lib/configurator.py
@@ -292,8 +292,6 @@ class Configurator(object):
         # if all of the widgets have been cleaned up correctly, this will delete the manager
         vtk_ui.manager.delete_manager(self.interactor)
 
-        self.interactor.Render()
-
     def release(self, object, event):
         if self.clicking is None:
             return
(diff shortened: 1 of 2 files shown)


Packages/vcs/Lib/vtk_ui/manager.py
```diff
@@ -78,6 +78,7 @@ def detach(self):
             w.detach()
         if self.window.HasRenderer(self.renderer):
             self.window.RemoveRenderer(self.renderer)
+        self.renderer.RemoveAllViewProps()
```
aashish24: You shouldn't need this but if you do then you want to set self.renderer = None since at this point I think it won't be of further use.

Packages/vcs/Lib/configurator.py
```diff
@@ -292,8 +292,6 @@ def detach(self):
         # if all of the widgets have been cleaned up correctly, this will delete the manager
         vtk_ui.manager.delete_manager(self.interactor)
 
-        self.interactor.Render()
```
aashish24: +1

Packages/vcs/Lib/vtk_ui/manager.py
chaosphere2112: Yeah, I basically only added this line as backup in case I screwed something up elsewhere; the "detach" functions are used throughout my `vtk_ui` package to make elements completely remove themselves from the VTK render stack and make them garbage collectable. I'll go ahead and set renderer to `None` here.

PR #1179, opened 2015-03-27
Fix ratio resizing issue

chaosphere2112:
Currently, when you resize a plot that has a ratio in `interact()` mode, things go badly.

```python
import vcs, cdms2
datafile = cdms2.open(vcs.prefix + "/sample_data/clt.nc")
clt = datafile("clt")

x = vcs.init()
x.plot(clt, ratio="autot")
x.interact()
```

Here's how the plot looks when you first make it (Perfectly correct)
screen shot 2015-03-27 at 11 57 35 am

Here's how it looks when you shrink the window (Not too bad...)
screen shot 2015-03-27 at 11 57 50 am

But hey, let's compare that to another plot that was actually plotted at that window size, for giggles...

```python
# press "Q", resume execution
y = vcs.init()
# You can plot whatever here, it doesn't matter initially
y.plot(clt)
# Just needed to get the window open...
y.clear()
# Resize to the size of x's window
y.backend.renWin.SetSize(x.backend.renWin.GetSize())
y.plot(clt, ratio="autot")
raw_input("Huh, would you look at that...")
```
Here's the x canvas' plot:
screen shot 2015-03-27 at 11 57 50 am

Here's the y canvas' plot:
screen shot 2015-03-27 at 11 58 00 am

While we can debate whether the labels should run over each other like that, the second behavior is the "correct" behavior, as the resizing in the first window only happens while in interact mode.

This PR should fix this issue– I looked into it, and plots given a ratio actually already generate a new template. My system was snagging that copied template, and then using it as the basis for a new template. This would cause some cascading error in the resizing logic, which is why the resize came out weird. Since the template is already a copy, I can just add the original and the copy into the existing store of templates. As a happy side effect, the code got a lot shorter and less wacky.

@doutriaux1 @aashish24 Mind taking a look?

### DIFF
--- a/Packages/vcs/Lib/configurator.py
+++ b/Packages/vcs/Lib/configurator.py
@@ -210,57 +210,21 @@ class Configurator(object):
                 self.init_toolbar()
 
         self.displays = [vcs.elements["display"][display] for display in self.canvas.display_names]
-
         for display in self.displays:
 
             if display._template_origin in self.templates:
-                # Adjust data coords if display.ratio is not none
-                if display.ratio is not None:
-                    t = vcs.gettemplate(display.template)
-                    orig = vcs.gettemplate(self.templates[display._template_origin])
-                    t.data._ratio  = -999.
-                    t.data._x1     = orig.data._x1
-                    t.data._x2     = orig.data._x2
-                    t.data._y1     = orig.data._y1
-                    t.data._y2     = orig.data._y2
-                    t.box1._x1     = orig.box1._x1
-                    t.box1._x2     = orig.box1._x2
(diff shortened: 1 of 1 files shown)


aashish24:
LGTM but I will wait for @doutriaux1 to review since I am not sure if this could open some other issues.

chaosphere2112:
Sounds good. It should only affect the interactive code; the "Save Template" and "Reset Template" buttons in the Configure Menu in particular, but I'm happy to wait till @doutriaux1 has time to check this out on Monday.

doutriaux1:
LGTM as well, it's much cleaner like this too!

Issue #1199, opened 2015-04-07
Cannot Animate Isoline


chaosphere2112:
Same basic issue as #1141, but with Isoline.

doutriaux1:
seems like a duplicte of #1194 closing. Please re-open if different.

doutriaux1:
actually will close 1194 rather than this one since the PR indicates this bug

PR #1200, opened 2015-04-07
Fix for #1199

chaosphere2112:
@doutriaux1 Check to make sure I did the right thing here; I was just emulating what you had done for Isofill


doutriaux1:
@chaosphere2112 it the right thing to do "except" labels do not get updated. I'm working on this do not worry about it.

PR #1203, opened 2015-04-08
Isoline anim labeled or not 1199

doutriaux1:
@chaosphere2112 this is yours plus labels animated as well
@dlonie @aashish24 probably won't hurt if you take a look at it as well


Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1824,11 +1825,28 @@ def update_input(self,vtkobjects,array1,array2=None,update=True):
                       mapper = missingMapper2
                       wrp = a[2]
                   else:
-                      mapper = vtk.vtkPolyDataMapper()
-                      mapper.SetInputConnection(ports[i].GetOutputPort())
-                      lut,rg = vtkobjects["vtk_backend_luts"][i]
-                      if lut is not None:
-                          mapper.SetLookupTable(lut)
+                      ## Labeled contours are a different king
```
dlonie: king?

dlonie:
LGTM if it fixes the issues.

Packages/vcs/Lib/VTKPlots.py
doutriaux1: kind I can't type ;)

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -913,13 +915,12 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
...
-
-              mapper.SetTextProperties(tprops)
          else:
             # No text properties specified. Use the default:
-            tprop = vtk.vtkTextProperty()
-            vcs2vtk.prepTextProperty(tprop, self.renWin.GetSize())
-            mapper.SetTextProperty(tprop)
+            tprops = vtk.vtkTextProperty()
+            vcs2vtk.prepTextProperty(tprops, self.renWin.GetSize())
+         mapper.SetTextProperty(tprops)
```
chaosphere2112: This line is wrong; should be indented. Breaks in test 254 and test 255.

chaosphere2112:
@doutriaux1 ctest failed on 254 and 255; complained that mapper.SetProperty wants a vtkTextProperty and received a vtkTextPropertyCollection

doutriaux1:
@chaosphere2112 ok all ctests pass now.

chaosphere2112:
:+1: All tests pass, animations work. Isoline labels make the animation *really* slow, though. I'm running a quick profile over it to see if there's anything obvious that could get optimized; I'll share the results of that with you before merging.

dlonie:
@chaosphere2112 Don't be surprised if its the labeling mapper that is eating up the extra time -- walking over and projecting the data into display coordinates (to find smooth regions to label) is pretty expensive!

chaosphere2112:
@dlonie Looks like we're doing 2 renders per frame; @doutriaux1 I'll see if I can find the extra render.

chaosphere2112:
D'oh, I was animating the slow way (making pngs). I'll try profiling the faster animation technique.

chaosphere2112:
ncalls|tottime|percall|cumtime|percall2|filename:lineno(function)
------|-------|-------|-------|--------|-------------------------
120|78.429|0.654|78.429|0.654|{built-in method Render}
240|4.091|0.017|4.091|0.017|{built-in method SetMapper}
600|2.506|0.004|2.506|0.004|{built-in method Update}
120|0.077|0.001|6.894|0.057|VTKPlots.py:1792(update_input)
120|0.03|0|0.03|0|{built-in method GetXRange}
120|0.025|0|7.151|0.06|VTKAnimate.py:13(update_input)
3840|0.024|0|0.031|0|cdmsobj.py:451(_listatts)
1800|0.02|0|0.038|0|axis.py:897(isCircular)
120|0.017|0|0.32|0.003|vcs2vtk.py:651(doWrap)
120|0.014|0|0.014|0|{built-in method SetInputData}
480|0.012|0|0.037|0|numeric.py:2156(allclose)
960|0.012|0|0.044|0|tvariable.py:84(_update_from)
360|0.011|0|0.011|0|{built-in method SetInputConnection}
1800|0.011|0|0.022|0|{method 'reduce' of 'numpy.ufunc' objects}
120|0.01|0|0.01|0|{built-in method AddInputData}

This is from 120 frames of animation using the "happy path". This was everything that had more than .01 seconds per call; I think it looks pretty reasonable. The happy path animation is still a lot slower than the other animations, but I think that @dlonie's explanation of the labels pretty much covers it. The only thing I see (from a black-box approach of not knowing what's going on under the covers) is that it does look like there may be twice as many calls to `SetMapper` as might be strictly necessary, though I don't think that'll rescue us too much time.

doutriaux1:
@dlonie @aashish24 @chaosphere2112 what is holding this one up?
It still passes all ctest (I even merged master in it before testing). See:
https://open.cdash.org/buildSummary.php?buildid=3780429

chaosphere2112:
@doutriaux1 The only thing that was holding it up was that the animations were quite slow, but there doesn't look like anything obvious that can be done to fix that. :+1: from me

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -913,13 +915,13 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
...
                   del(vcs.elements["texttable"][tt])
-
-              mapper.SetTextProperties(tprops)
+          mapper.SetTextProperties(tprops)
          else:
             # No text properties specified. Use the default:
-            tprop = vtk.vtkTextProperty()
-            vcs2vtk.prepTextProperty(tprop, self.renWin.GetSize())
-            mapper.SetTextProperty(tprop)
+            tprops = vtk.vtkTextProperty()
```
dlonie: This variable name should to be changed back -- it's only a single text property object.

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1824,11 +1826,28 @@ def update_input(self,vtkobjects,array1,array2=None,update=True):
...
+                              mapper.SetLookupTable(lut)
+                          else:
+                              stripper = vtk.vtkStripper()
+                              stripper.SetInputConnection(ports[i].GetOutputPort())
+                              mapper.SetInputConnection(stripper.GetOutputPort())
+                              stripper.Update()
+                              tprops = vtkobjects["vtk_backend_contours_labels_text_properties"]
+                              mapper.GetPolyDataMapper().SetLookupTable(lut)
+                              mapper.SetLabelVisibility(1)
+                              mapper.SetTextProperty(tprops)
```
dlonie: This won't work in all cases. If text colors/styles are specified on a per-level basis, `tprops` will be a `vtkTextPropertyCollection`, not a `vtkTextProperty`, and will need to be passed to `vtkLabeledContourMapper::SetTextProperties(vtkTextPropertyCollection*)`, rather than `vtkLabeledContourMapper::SetTextProperty(vtkTextProperty*)`. See http://www.vtk.org/doc/nightly/html/classvtkLabeledContourMapper.html for details.

doutriaux1: nice catch! Do we have a way to test animations? Short of creating it saving the pngs and making sure they all match the baseline? It would be nice to setup test case.

dlonie: What we've done in some of the ParaView tests is to run an animation from start to finish and check certain key frames (usually just the first and the last). It seems to work rather well at catching errors.

VCS should be able to do something similar, AIUI. We'd just need to call the image comparison code at multiple points in the test.

doutriaux1:
@dlonie while addressing the issue you brought up (textproperties vs textproperty) I did 3 things
1- Made sure isolines actually are of the color asked by user (they were all black before no matter what)
2- added tests (baseline pr: https://github.com/UV-CDAT/uvcdat-testdata/pull/36 )
3- realized the text colors seem to be messed up if you look at the test baselines: https://raw.githubusercontent.com/UV-CDAT/uvcdat-testdata/master/baselines/vcs/test_isoline_labels.png or https://github.com/UV-CDAT/uvcdat-testdata/blob/master/baselines/vcs/test_vcs_isoline_labels_multi_label_input_types.png

If you look at the pngs carefully you will realize that the text colors change for isolines of same values (look at 50 or 80 for example in the second png) this is wrong, all isolines of the same value should have the same color.

Also I think the code will need more tweaks beacuse right now we are using the same "lut" for both isolines and textcolors but it is conceivable that the user ask for different colors (isolines in shades of blue, but text in shades of red for readibilty)

I think you should address this in a separate PR so we don't delay this PR too much longer, but it is high priority I will create an issue .

Thanks,

dlonie:
@doutriaux1 Well that's definitely not right! I'll pull and test this branch then take a look at the color issues.

testing/vcs/test_animate_isoline.py
```diff
@@ -0,0 +1,33 @@
...
+x.drawlogooff()
+x.setbgoutputdimensions(1200,1091,units="pixels")
+
+iso=x.createisoline()
+x.plot(s,iso,bg=1)
+x.animate.create()
+print "Saving now"
+prefix= os.path.split(__file__)[1][:-3]
+x.animate.save("%s.mp4"%prefix)
+pngs = x.animate.close(preserve_pngs = True) # so we can look at them again
```
dlonie: @doutriaux1 @chaosphere2112 Will this use a unique prefix for the png files?

Dashboards often run `ctest -jX` to speed things up by running test in parallel. If these animation tests would overwrite each other, we'll need to make sure that parallel testing is disabled.

testing/vcs/test_animate_isoline.py
```diff
@@ -0,0 +1,33 @@
...
+src_pth = sys.argv[1]
+pth = os.path.join(src_pth,prefix)
+ret = 0
+for p in pngs:
+  print "Checking:",p
+  ret += checkimage.check_result_image(p,os.path.join(pth,os.path.split(p)[1]),checkimage.defaultThreshold)
```
dlonie: Very cool, glad you got animation testing working! My one suggestion is to limit the number of comparisons to every fourth frame or so. It should still catch any failures, but will be less of a headache when the baselines need updating ;-)

testing/vcs/test_animate_isoline.py
chaosphere2112: @dlonie Yup, all animation PNGs are generated with a unique prefix.

dlonie: Cool, just wanted to be sure!

testing/vcs/test_animate_isoline.py
dlonie: Feel free to disagree, btw. I don't object to merging this as-is, just a suggestion.

dlonie:
Fixed #1230 in this branch.

This change requires UV-CDAT/VTK#3

aashish24:
LGTM :+1: all tests are passing.

PR #1207, opened 2015-04-08
Fix #1197

### LINKED ISSUE #1197: clear canvas while animation running in GUI doesn't seem to stop it
@williams13 showed me and he keeps getting vtk warn messages after clearing canvas as if animation was still running in background

chaosphere2112:
Fix for #1197; adds a test that makes sure the configurator is cleaned up when the canvas is closed / the GUI kills a canvas.


doutriaux1:
works for me on ubunu (except the vtk_ui test that seg fault but you know about this)

Issue #1198, opened 2015-04-07
boxfill custom in GUI


doutriaux1:
plot clt in gui, click edit pencil, select "custom" boxfill type, create new min max and generate ranges, click "Apply" the plot disappear, now move mouse over to "open configure" widget, plot reappears

doutriaux1:
also legend does not appear

chaosphere2112:
@doutriaux1 I'm not entirely clear on this one, you'll have to help me reproduce tomorrow.

chaosphere2112:
This is triggered by a couple of errors that happen during the update; 

```
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/core/modules/vistrails_module.py", line 327, in update
    self.compute()
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/packages/uvcdat_cdms/init.py", line 1282, in compute
    self.cellWidget = self.displayAndWait(QCDATWidget, input_ports)
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/packages/spreadsheet/basic_widgets.py", line 247, in displayAndWait
    return spreadsheetWindow.displayCellEvent(e)
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/packages/spreadsheet/spreadsheet_window.py", line 490, in displayCellEvent
    sheet.setCellByType(row, col, e.cellType, e.inputPorts)
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/packages/spreadsheet/spreadsheet_tab.py", line 326, in setCellByType
    oldCell.updateContents(inputPorts)
  File "/Users/fries2/projects/dotfiles/bin/uvcdat/2.2-3/vistrails/vistrails/packages/uvcdat_cdms/init.py", line 1457, in updateContents
    self.canvas.plot(cgm,*args,**kwargs)
  File "/Users/fries2/bin/uvcdat/2.2-3/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.py", line 2613, in plot
    a = self.__plot( arglist, keyargs )
[...]

PR #1210, opened 2015-04-09
Add test which replicates Dean's boxfill update issue.

dlonie:
This test triggers a reproducable failure when changing levels on a boxfill plot.

*DO NOT MERGE* This just replicates the bug. It does not provide a fix nor a valid baseline.

Refs #1198.


dlonie:
@aashish24 Here's the script you were looking for.

@doutriaux1 @chaosphere2112 This reproduces #1198.

aashish24:
Nice! this is exactly what we need to address our issues!! thank you for doing it. @williams13 please check it out. Thanks @dlonie

williams13:
What exactly am I looking for?

aashish24:
@williams13 please have a look at the test code. These are the kinds of test we need to test out state changes working.

doutriaux1:
@aashish24 @dlonie @chaosphere2112 @williams13 this branch is now ready to merge
it fixes the issue adds a bunch of tests for it and also fix/tweaks the way we were setting extensions on graphics methods. Before they were possibilities for the use rto pass values tat would not trigger the right calls.

doutriaux1:
goes with:
https://github.com/UV-CDAT/uvcdat-testdata/pull/33

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -817,12 +817,14 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
           cot.SetInputData(vtk_backend_grid)
 
       levs = gm.levels
-      if (isinstance(gm,isoline.Gi) and numpy.allclose( levs[0],[0.,1.e20])) or numpy.allclose(levs,1.e20):
-        levs = vcs.mkscale(mn,mx)
-        if len(levs)==1: # constant value ?
-          levs = [levs[0],levs[0]+.00001]
-        Ncolors = len(levs)
-        if isinstance(gm,(isofill.Gfi,meshfill.Gfm)):
+      ## Apparently in some cases
```
dlonie: ?

dlonie:
New tests pass, LGTM.

Packages/vcs/Lib/VTKPlots.py
aashish24: Can we remove this line or explain it better?

Packages/vcs/Lib/utils.py
```diff
@@ -571,15 +572,15 @@ def scriptrun(script):
...
-            except Exception,err:
-              print "failed",typ,nm,err
-    except Exception,err: #ok could not read json file maybe it is an old initial.attributes
-      if os.path.split(script)[-1] == "initial.attributes":
-        _scriptrun(script)
-      else:
-        warnings.warn("unable to source file: %si %s" % (script,err))
+            #except Exception,err:
+            #  print "failed",typ,nm,err
+    #except Exception,err: #ok could not read json file maybe it is an old initial.attributes
```
aashish24: If code is not useful anymore, I would suggest we remove it. Un commenting a lot of code is not desired.

Packages/vcs/Lib/VTKPlots.py
doutriaux1: Looks like old code not completely removed, will remove in a separate PR

Packages/vcs/Lib/utils.py
doutriaux1: it is actually useful, was commented out while debugging to better catch error message/traceback. Will submit a new PR right now to put it back in

doutriaux1: @aashish24 it's in https://github.com/UV-CDAT/uvcdat/tree/boxfill_bug_branch_clean_up
But there is still an issue with boxfill_bug in GUI. I might need to tweak this branch some more, so I'm not issuing the PR yet.

aashish24: thanks. just wanted to make sure. Like I said if code is not useful remove it. Commenting a big chunk of code makes code unreadable and non-production like.

Issue #1208, opened 2015-04-08
test_vtk_ui_button* segfaults on Ubuntu


chaosphere2112:
Looks like there's a segfault in the VTK UI tests on ubuntu; I'll sort this out tomorrow.
```
Test project /home/fries2/build
      Start 272: test_vtk_ui_button_draw
 1/11 Test #272: test_vtk_ui_button_draw ............***Exception: SegFault  2.98 sec
      Start 273: test_vtk_ui_button_action
 2/11 Test #273: test_vtk_ui_button_action ..........***Exception: SegFault  1.08 sec
      Start 274: test_vtk_ui_button_corner_radius
 3/11 Test #274: test_vtk_ui_button_corner_radius ...***Exception: SegFault  1.02 sec
      Start 275: test_vtk_ui_button_dimensions
 4/11 Test #275: test_vtk_ui_button_dimensions ......***Exception: SegFault  1.05 sec
      Start 276: test_vtk_ui_button_states
 5/11 Test #276: test_vtk_ui_button_states ..........***Exception: SegFault  1.06 sec
      Start 277: test_vtk_ui_button_click
 6/11 Test #277: test_vtk_ui_button_click ...........***Exception: SegFault  1.07 sec
      Start 278: test_vtk_ui_button_font
 7/11 Test #278: test_vtk_ui_button_font ............***Exception: SegFault  1.09 sec
      Start 279: test_vtk_ui_button_alignment
 8/11 Test #279: test_vtk_ui_button_alignment .......***Exception: SegFault  1.12 sec
      Start 280: test_vtk_ui_button_image
 9/11 Test #280: test_vtk_ui_button_image ...........***Exception: SegFault  1.14 sec
      Start 281: test_vtk_ui_button_tooltip_show
10/11 Test #281: test_vtk_ui_button_tooltip_show ....***Exception: SegFault  1.09 sec
      Start 282: test_vtk_ui_button_tooltip_hide
11/11 Test #282: test_vtk_ui_button_tooltip_hide ....***Exception: SegFault  1.17 sec

0% tests passed, 11 tests failed out of 11

Total Test time (real) =  13.92 sec

The following tests FAILED:
	272 - test_vtk_ui_button_draw (SEGFAULT)
	273 - test_vtk_ui_button_action (SEGFAULT)
	274 - test_vtk_ui_button_corner_radius (SEGFAULT)
	275 - test_vtk_ui_button_dimensions (SEGFAULT)
[...]

PR #1211, opened 2015-04-09
VTK_UI Tests no longer segfault on Ubuntu (and other X based systems)

chaosphere2112:
Fix for #1208; I was calling `CreateOneShotTimer` on an uninitialized `vtkXRenderWindowInteractor`, which led to segfaults on Ubuntu (and presumably other X-based systems). Now the `InterfaceManager` will check if the interactor is initialized before trying that, and will jut call render on the RenderWindow if it is not.


doutriaux1:
works for my on my ubunutu

PR #1229, opened 2015-04-22
Vtk refactor

dlonie:
Refactor VTKPlots.plot2D.

@doutriaux1 @chaosphere2112 Please read the following, and test/review the patch. I'll be asking specific questions inline, as there are some conventions used that I couldn't quite figure out from just reading the code/comments.

This change should neither affect nor improve external behavior at this point, but just organizes the code and separates the logic for different plot types to make the implementations more robust to changes and easier to read/follow/modify/debug.

The Pipeline class is an interface for interacting with VTK. At this point, it defines a single virtual method, plot(...), which takes arguments similar to VTKPlots.plot(...).

Eventually, the interface can grow to include methods for e.g. animation update, but for now it still just returns the map of objects, so that code should continue to function as before.

This patch:

- Creates initial Pipeline interface.
- Add shared logic from Plot2D into Pipeline2D subclass.
- Adds specific implementations as Boxfill-, Isoline-, Isofill-, and MeshfillPipeline subclasses.
- **Cleans up style to conform to PEP 8. I will be enforcing this in future patches to the vcsvtk submodule, so please become familiar with the conventions:** https://www.python.org/dev/peps/pep-0008/
- **Creates vcs2vtk.isTrue(...) to perform consistent user_input-->bool conversions.**
-- Sometimes we check if a variable is true with `if var == 'y'`, sometimes with `if var == 1`, sometimes with `if var in ('y', 1, True)`, and sometimes with just `if var:`. Let's start using this function to keep things consistent and working right the first time. This also adds the advantage of being able to fix bugs more easily -- for example, if all non-zero integers should mean true (rather than just `1`), we can easily update all usages by just changing the function's definition.

Feel free to ask questions or make suggestions if there's something you don't like.

**This will break several pending branches on UV-CDAT.** Ping me on them if you want help porting them to the new framework after this topic lands in master.


Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -15,6 +15,14 @@
 f = open(os.path.join(vcs.prefix,"share","vcs","wmo_symbols.json"))
 wmo = json.load(f)
 
+def isTrue(x):
+    """Test if an unsanitized user input is 'true'."""
+    # I'm just encapsulating existing logic here, but this seems to miss some
+    # other cases users may enter. What about:
+    #    - Non-zero integers other than 1?
+    #    - "yes", "Yes", "Y"?
+    return x in ("y", 1, True)
```
dlonie: @doutriaux1 Should we be allowing these other cases to mean 'true'? They're fairly standard conventions, but I'm not sure what VCS expects. Is this information in the documentation?

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+                if numpy.allclose(levs2[0], 1.e20):
+                    levs2[0] = 0
+                for i in range(len(levs2) - 1):
+                    self._contourLevels.append([levs2[i], levs2[i+1]])
+
+        # Contour colors:
+        self._contourColors = self._gm.fillareacolors
+        if self._contourColors is None:
+            # TODO BUG levs2 may not be defined here...
+            self._contourColors = vcs.getcolors(levs2, split=0)
```
dlonie: After simplifying the contour level logic, this bug became more obvious. I'll put in a test/bug report soon.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+            # TODO Charles, why would mapper be a list? The custom boxfill code
+            # pushes single-element lists into the mapper list, though a comment
+            # suggests that this means a textured pattern should be used. I'll
+            # clean this up to be more clear, but need to understand the current
+            # usage pattern first.
+            if isinstance(mapper, list):
+                act.SetMapper(mapper[0])
+            else:
+                mapper.Update()
+                act.SetMapper(mapper)
```
dlonie: @doutriaux1 This was a common question I had when reading this code. The `mappers` variable would have a mix of `vtkMapper` objects and single-element lists containing a `vtkMapper`, and I can't figure out the difference or why this would be. Comments for these sort of non-obvious conventions help immensely ;)

So, what is the difference between `mapper` and `[mapper,]` in the `mappers` list?

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+                act = vcs2vtk.doWrap(act, [x1,x2,y1,y2], self._dataWrapRange)
+
+            # TODO Charles, what's the difference between these two branches?
+            # Why not just store the actor and use GetMapper() to get the mapper
+            # when needed? The distinction isn't very clear to other developers
+            # that need to read the code.
+            if mapper is self._maskedDataMapper:
+                actors.append([act, self._maskedDataMapper, [x1,x2,y1,y2]])
+            else:
+                actors.append([act, [x1,x2,y1,y2]])
```
dlonie: @doutriaux1 Another non-obvious change in behavior. Does conditional branch need to exist, since the mapper is accessible through `act.GetMapper()`?

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+
+            # TODO Charles, the custom boxfill code *will* trigger this block to
+            # execute, but without mapper[1] defined. Why are some entries in
+            # mappers actual mappers while others are single-element lists
+            # containing a mapper?
+            if isinstance(mapper, list):
+                ## This is the sport to add patterns
+                #act.GetMapper().ScalarVisibilityOff()
+                #act.SetTexture(mapper[1])
+                pass
```
dlonie: @doutriaux1 This code is commented out, but seems to be tied into the "mappers contains mappers and sometimes lists" issue. Note that none of the implementations would produce an entry in mappers that would have a second element (`mapper[1]` would throw a bounds error).

Packages/vcs/Lib/vcsvtk/isofillpipeline.py
```diff
@@ -0,0 +1,296 @@
...
+            # it, tests start failing as vtkCleanUnstructuredGrid is a
+            # ParaView filter that is not currently available in VTK.
+            # There haven't been any problems from skipping this step in the
+            # tests, is it really needed?
+            if self._vtkDataSet.IsA("vtkUntructuredGrid"):
+                cln = vtk.vtkCleanUnstructuredGrid()
+                cln.SetInputConnection(c2p.GetOutputPort())
+                self._vtkPolyDataFilter.SetInputConnection(cln.GetOutputPort())
+            else:
+                self._vtkPolyDataFilter.SetInputConnection(c2p.GetOutputPort())
```
dlonie: @doutriaux1 See the comment/typo above. The true branch of this conditional has never executed since there is no class in VTK matching the `IsA` argument. This hasn't cause problems so far, any objections to me removing this check and just leaving the `else` body?

Packages/vcs/Lib/vcsvtk/isofillpipeline.py
```diff
@@ -0,0 +1,296 @@
...
+        if len(cots) > 0:
+            self._resultDict["vtk_backend_contours"] = cots
+        if len(geos) > 0:
+            self._resultDict["vtk_backend_geofilters"] = geos
+
+        numLevels = len(self._contourLevels)
+        if mappers == []: # ok didn't need to have special banded contours
+            # TODO Charles, this branch will never execute, unless numLevels==0.
+            # Is that a supported usecase, or should that throw an error?
+            mapper = vtk.vtkPolyDataMapper()
```
dlonie: @doutriaux1 For isofills, `mappers` will never be empty as long as there is at least one contour level used. Are level-less isofills considered valid in VCS? If not, I can cut this conditional branch out.

Packages/vcs/Lib/vcsvtk/isolinepipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        actors = []
+        for mapper in mappers:
+            act = vtk.vtkActor()
+            # TODO Charles -- Mapper is always a list, what is
+            # this supposed to mean?
+            if isinstance(mapper, list):
+                act.SetMapper(mapper[0])
+            else:
+                mapper.Update()
+                act.SetMapper(mapper)
```
dlonie: @doutriaux1 For isolines, all entries in `mappers` are single-element lists with a single mapper. Is there a reason for the extra level of list nesting?

Packages/vcs/Lib/vcsvtk/meshfillpipeline.py
```diff
@@ -0,0 +1,305 @@
...
+              mapper.SetLookupTable(lut)
+              mapper.SetScalarRange(l[j], l[j+1])
+              luts.append([lut, [l[j], l[j+1], False]])
+              ## Store the mapper only if it's worth it?
+              ## Need to do it with the whole slab min/max for animation
+              ## purposes
+              if not (l[j+1] < wholeDataMin or l[j] > wholeDataMax):
+                  # TODO Charles: Why is this appending a single element list,
+                  # instead of just an object?
+                  mappers.append([mapper,])
```
dlonie: @doutriaux1 Another question about list nesting. Is this needed, or just unnecessary complexity?

Packages/vcs/Lib/vcsvtk/pipeline2d.py
```diff
@@ -0,0 +1,194 @@
...
+        self._data1 = None
+        self._data2 = None
+        self._contourLevels = None
+        self._contourColors = None
+        self._vtkDataSet = None
+        self._vtkGeoTransform = None
+        self._vtkDataSetBounds = None
+        self._colorMap = None
+        self._useContinents = None
+        self._dataWrapRange = None # TODO Need some clarification on what this does.
```
dlonie: @doutriaux1 This member holds the `wrap` value returned from the vcs2vtk.genGrid[...]()` methods, for which no documentation exists. Can you provide a brief description of what this variable represents and its expected format so I can update this class's documentation?

Packages/vcs/Lib/vcsvtk/pipelinefactory.py
```diff
@@ -0,0 +1,24 @@
+def createPipeline(graphics_method, context):
+    """Create and initialize a Pipeline subclass from a graphics method.
+
+    Returns None if the graphics method is not recognized.
+    """
+    # TODO Charles, is this the best way to identify a graphics method? There is
+    # no common superclass for them.
+    if not hasattr(graphics_method, 'g_name'):
+        return None
```
dlonie: @doutriaux1 Is there a better way to identify a graphic method object?

Packages/vcs/Lib/vcsvtk/pipeline1d.py
```diff
@@ -0,0 +1,123 @@
...
+        if hasattr(data1, "_yname"):
+          del(data1._yname)
+        del(vcs.elements["line"][l.name])
+        if gm.marker is not None:
+            del(vcs.elements["marker"][m.name])
+
+        if tmpl.legend.priority > 0:
+            legd = self._context.canvas.createline()
+            legd.x = [tmpl.legend.x1, tmpl.legend.x2]
+            legd.y = [tmpl.legend.y1, tmpl.legend.y1] # TODO Charles, is this a typo?
```
dlonie: @doutriaux1 Just checking, should this be `[y1, y2]` instead of `[y1, y1]`?

aashish24:
@doutriaux1 will provide comments after we tag 2.2 today. @dlonie we would like to have it merged as soon as possible.

dlonie:
Any progress on this?

doutriaux1:
Looks like this is the week @dlonie

Packages/vcs/Lib/vcs2vtk.py
doutriaux1: @dlonie most attribute setting go through VCSValidationFunction which takes any of the has an input and sets it to True anyway. I know of one case where somehow VisTrails bypass this and sets directly the _attribute and we endup with "False" (string not boolean) but this wouldn't cover it anyway. I think we should remove this test, and if anything breaks, make sure the attribute is actually set via VCSValidation function.

doutriaux1: as far as doc goes if we use VCSValidation functions, then a "bad" value triggers an error that list "correct" possible values.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+import numpy
+import vcs
+import vtk
+import warnings
+
+class BoxfillPipeline(Pipeline2D):
+    """Implementation of the Pipeline interface for VCS boxfill plots.
+
+    Internal variables:
+        - self._legend: Contour labels.
```
doutriaux1: legend actually has another meaning in VCS vs "labels" the legend is used to display which lines represent what. If think "labels" or "ContourLabels" might be better here.

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -446,8 +447,11 @@ def plot(self,data1,data2,template,gtype,gname,bg,*args,**kargs):
...
     vtk_backend_grid = kargs.get("vtk_backend_grid",None)
     vtk_backend_geo = kargs.get("vtk_backend_geo",None)
-    if gtype in ["boxfill","meshfill","isofill","isoline"]:
-      returned.update(self.plot2D(data1,data2,tpl,gm,vtk_backend_grid=vtk_backend_grid,vtk_backend_geo=vtk_backend_geo))
+
+    pipeline = vcsvtk.createPipeline(gm, self)
+    if pipeline is not None:
+        returned.update(pipeline.plot(data1, data2, tpl, gm,
+                                      vtk_backend_grid, vtk_backend_geo))
     elif gtype in ["3d_scalar", "3d_dual_scalar", "3d_vector"]:
```
doutriaux1: why not 3D stuff?

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
doutriaux1: @dlonie I think originally we only had 1 mapper, but then some methods required lists of mappers (like isofill) so it became a list, and in the case of single mapper i made it a single element list in order to have a consistent post-process where we do fiToViewPort

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
doutriaux1: probably didn't know about GetMapper when I started developing this. But also I think the actors get replaced when we do fitToViewPort, so that's probably why I opted to save the mappers, in any case you're right we should use GetMapper

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
doutriaux1: @dlonie ok this is a place holder for when we actually put in patterns, as I understood it back then, the "texture" is actually a mapper. So for each level I needed 2 mappers, the "level" mapper (color etc...) and the texture mapper, hence the list.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -0,0 +1,383 @@
...
+            if self._gm.boxfill_type == "log10":
+                for k in self._legend.keys():
+                    self._legend[float(numpy.ma.log10(self._legend[k]))] = \
+                          self._legend[k]
+                    del(self._legend[k])
+
+        # Use consecutive colors:
+        self._contourColors = range(self._gm.color_1, self._gm.color_2 + 1)
+
+    def _updateContourLevelsAndColorsForCustomBoxfill(self):
```
doutriaux1: this bit is probably reusable for isofill. Shouldn't we create a "levels" class and inherit from this, so that we don't need to duplicate efforts between boxfill custon/isofill/meshfill

Packages/vcs/Lib/vcsvtk/isofillpipeline.py
doutriaux1: I trust you, anyway I think it's either unstructured or structured, can't think of anything else right now. So "else" is most likely ok.

Packages/vcs/Lib/vcsvtk/isolinepipeline.py
doutriaux1: @dlonie the second mapper was supposed to be for texture patterns, we don't need it for isoline.

Packages/vcs/Lib/vcsvtk/meshfillpipeline.py
doutriaux1: @dlonie needed here for textures

doutriaux1: @dlonie again there is a LOT of redundancy between boxfill_custom/isofill/meshfill. We probably want to create a common class to inherit from for all these. no?

Packages/vcs/Lib/vcsvtk/pipeline1d.py
doutriaux1: @dlonie nope we're drawing a straight horizontal line

Packages/vcs/Lib/vcsvtk/pipeline2d.py
doutriaux1: wrap is for "wrap around" i.e if data are 0/360 but user request -180/180 the "wrrp" would tell us it is ok to wrap the data, I guess we could rename this "modulo", keep in mind meshfill has a "wrap" attribute, but we could dperecate it in favor of modulo. basically it indicates the module in y,x directions wrap=[0,360] means no modulo on Y 360 modulo on X direction

Packages/vcs/Lib/vcsvtk/pipelinefactory.py
doutriaux1: we should use vcs.isgraphicsmethod()

testing/vcs/test_boxfill_number_color_more_than_number_levels.py
```diff
@@ -34,6 +34,7 @@
     failed = True
 except:
     failed = False
+    raise
```
doutriaux1: probably shoud be
```
     raise RuntimeError("This test did not issue warning as expected")
```

doutriaux1:
@dlonie I'm going to test it a bit now that I have commented on your questions. LOTS of conflicts with master/release now... Are you going to bring everything back in?
@aashish24 I would like to merge this soon, but not in master until 2.2 is out. Where should this be merged though? It would be nice to start new development from this now.

dlonie:
@doutriaux1 Don't worry too much about testing or resolving conficts at this point. I'll merge the changes in master back into this, and then we can do testing. After I get this resync'd with master, let's merge ASAP so we don't diverge too much again.

doutriaux1:
@dlonie ok, but where to merge to? There's conflict with master and I don't think we want master to get this right now as it has not been thoroughly tested. release is not an option, @aashish24 is it time to create a "devel"/"experimental" branch (off off master/release) ?

dlonie:
Definitely not release. I'd say testing and merging to master should be sufficient. There isn't any new behavior or changes to functionality here, all I've done is separate out the implementations and strip out the bits that would never execute for a given plot type.

I've specifically kept significant changes out of this branch to preserve the behavior of master to make sure that we can get it in quickly. I'd rather not put this into a long-term branch off of master, as that is going to just keep diverging and require manual updating.

Let's get this into master and then we can work on this going forward. `VTKPlots.plot2d` really needs to go away ASAP before any more bugs spawn from it :P

doutriaux1:
@dlonie fine by me but @aashish24 didn't even want the release bits to go into master... I'm not quite sure what he has in mind for master... Can you please merge master AND release into this one (or a new branch) for now so that we can really test it. I really don't want master to get dirty that's what people clone right out of git. Thanks.

dlonie:
Hmm. I'll just do master at the moment, and release when we merge release in. I wasn't aware that release was diverging from master like that. Typically changes that go into a release branch also go into master. @aashish24 What's going on with master?

testing/vcs/test_boxfill_number_color_more_than_number_levels.py
dlonie: I'd prefer leaving the blank `raise`, as this reraises the previous exception and preserves the stack trace of the actual error.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
dlonie: That's a possibility. As this point, I'm just separating the pieces out to see what we have. I'll start consolidating/cleaning them together as it makes sense later on.

Packages/vcs/Lib/VTKPlots.py
dlonie: No real reason - this can be moved later. My priority at this point was to untangle `plot2D`, and I did the vector/1d plots as well since they were easy. At some point all of these should go into pipeline objects.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
dlonie: Ah, this is the variable named `legend` in the original `plot2D`, so I just kept the name. I'll rename it.

doutriaux1:
@dlonie @aashish24 tried to explained it to me there: https://github.com/UV-CDAT/uvcdat/pull/1259

Packages/vcs/Lib/vcs2vtk.py
dlonie: That sounds far more preferable to me. I seem to recall a lot of patches going in that would add explicit checks for these cases, but if all of that is handled by the validator, then awesome! I'll change these to just check if the variable is True and any issues that pop up will be treated a validation bugs.

dlonie: Looks like this will fail after just checking for True:

https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/VTKPlots.py#L889

dlonie: @doutriaux1 Yep, treating that as True/False causes isolabels to always be drawn.

Can you point me at the code that should be validating this? I took a look through VCS_validation_functions.py and didn't see anything related to isoline labels.

dlonie: This spot too:

https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/VTKPlots.py#L834

If inputs are sanitized, we shouldn't need special cases for backwards compatibility, right?

doutriaux1: @dlonie labels are checked at:
https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/isoline.py#L540
Looks like it is speciafically checking  for "Yes/No", I think it's ok to check for Yes/No/True/False
LEt's replace with:
https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/VCS_validation_functions.py#L360
And mke it set to True/False rather than "y"/"n'

doutriaux1: @dlonie you're right let's assume it is sanatize at nput time. If something breaks it's better anyway to fix the part where it should have been sanitized

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
doutriaux1: oh I see we're in boxfill... yes this is probably the strings/labels to draw on the legend, user-specified. I guess it's an ok name in this case, since we're talking about labels on legend....

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
doutriaux1: :+1:

Packages/vcs/Lib/vcs2vtk.py
dlonie: Ah, thanks -- just what I was looking for.

Right now we have:

`checkTrueFalse`
`checkOnOff`
`checkYesNo`

Would it make sense to replace all of these with a standard `checkBoolean` that tests all of these cases and sets to `True`/`False`?

dlonie: Not as part of this patch, but as a later issue? It seems these all do the same thing but in slightly different ways.

Packages/vcs/Lib/vcsvtk/isofillpipeline.py
doutriaux1:
```
levels =  ([1.0000000200408773e+20, 1.0000000200408773e+20],)
```
is the closest thing from level-less it means "I trust you computer, give me your best shot"

testing/vcs/test_boxfill_number_color_more_than_number_levels.py
doutriaux1: good point.  I don't see the whole file right here but do we still need the line 39/40?

dlonie: Nope -- I'll take that out.

dlonie: Actually, we do. In case the warning exception isn't thrown, but no others are, we'll need that to catch it.

Packages/vcs/Lib/vcs2vtk.py
doutriaux1: Let's not touch these functions per say, we might need these for something else that really just need yesno (and not boolean) But you're right let's create an issue to create a unifying function that understands all these and sets to True/False, checkBoolean should really only check if the input is a boolean, let's think of a better name, I liked checkOnOff but that's already used.Maybe checkTrigger ?

testing/vcs/test_boxfill_number_color_more_than_number_levels.py
doutriaux1: yes... and that's actually the whole point of the test :smile:

Packages/vcs/Lib/vcs2vtk.py
dlonie: I'll add that when I make this change, and leave the other function as-is. We can always port usages of them later as needed.

I'll add a `checkBoolean` that strictly checks `True`/`False`, and a `checkFuzzyBoolean` that permits the other representations.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
dlonie: Ah, ok. I'll clean these up to just put mappers into the list as needed and remove the list-of-lists checks.

FYI, looking at the isofill case, that can be rewritten to use a single polydata/mapper, similar to the new scalar bar implementation. I'll leave it as-is for now and we can revisit it if it starts causing noticeable performance issues.

doutriaux1: @dlonie the list of list is really for the not-yet-implemented patterns bit. So yes for now we can take it out

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
dlonie: I'll leave this as-is for now with a TODO. It can be cleaned up when we start handling the animation updates in these classes.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
dlonie: Ok, in that case I'll take this out for now. We can re-evaluate that approach when adding pattern support.

Packages/vcs/Lib/vcsvtk/meshfillpipeline.py
dlonie: Agreed!

BTW, there is a common baseline class, Pipeline2D, and while there are a lot of _similar_ portions, many had tons of conditionals to implement different behaviors for the different backends. Pipeline2D has some of the more easily abstracted bits taken out already, and as the refactoring progresses I expect to put more of them in the base class.

Packages/vcs/Lib/vcsvtk/pipeline1d.py
dlonie: Okeydoke.

Packages/vcs/Lib/vcsvtk/pipeline2d.py
dlonie: Updated the documentation to:

`- _dataWrapModulo: Wrap modulo as [YMax, XMax], in degrees. 0 mean 'no wrapping'.`

dlonie:
Ok, new version is rebased on current master and incorporates feedback from @doutriaux1 :+1: Should be ready to merge...somewhere, as soon as we figure out where that somewhere is!

I still believe this should go into master. We don't want to merge this into release, as it shouldn't go into 2.2 -- it should go into the master branch as that's the target for non-release development. At least, this is how it typically works in other projects.

Otherwise, we're essentially freezing master while we stabilize the release, which more or less defeats the purpose of the release branch.

@aashish24 Can you clarify what goes where?

aashish24:
this change should not go into release. This is a major change and should only be merged into master. Earlier, I have merged relevant changes into master from release to make sure that we have all the bug fixes in master as well. After this branch goes into master, any changes to these files will have to target particular branch. So in summary: 

1) Merge this branch into master only if @doutriaux1 approved it
2) Merge any relevant changes from release into master (I can do it) 
3) Do not merge master into release anymore 
4) We should plan from 2.3 from master in next month or so.

dlonie:
@aashish24 Thanks, sounds good.

doutriaux1:
@aashish24 ok we all agree it should go into master only.
@aashish24 is everything from release into master as well? I would like to create a release branch on uvcdat-testdata before things get too crazy.

doutriaux1:
ok all test pass for me let's do this!
https://open.cdash.org/buildSummary.php?buildid=3801982

aashish24:
thanks @doutriaux1. @dlonie if we setup a mechanism to follow branch when we checkout testdata would that break anything? 

@doutriaux1 I will make sure that everything in release is in master as well.

doutriaux1:
thanks @aashish24

PR #1236, opened 2015-04-27
Issue 1234 boxfill anim broken

doutriaux1:
@aashish24 @dlonie @chaosphere2112 goes with:
https://github.com/UV-CDAT/uvcdat-testdata/pull/37

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1153,7 +1153,7 @@ class VTKVCSBackend(object):
       else:
         lmx= levs[-1]
       mapper.SetScalarRange(lmn,lmx)
-      returned["vtk_backend_luts"]=[[lut,[lmn,lmx,False]],]
+      returned["vtk_backend_luts"]=[[lut,[lmn,lmx,True]],]
 
     if missingMapper is not None:
       if isinstance(gm,meshfill.Gfm):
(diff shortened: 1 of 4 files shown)


Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1153,7 +1153,7 @@ def plot2D(self,data1,data2,tmpl,gm,vtk_backend_grid=None,vtk_backend_geo=None):
       else:
         lmx= levs[-1]
       mapper.SetScalarRange(lmn,lmx)
-      returned["vtk_backend_luts"]=[[lut,[lmn,lmx,False]],]
+      returned["vtk_backend_luts"]=[[lut,[lmn,lmx,True]],]
```
dlonie: These map entries really need some documentation. I have no idea what this change should do.

dlonie:
Building/testing now.

aashish24:
hey guys.. not picking on anyone but please use better short messages for the commit.

aashish24:
@dlonie I am assuming that you will have some feedback on it.

dlonie:
Yep, still building. Assuming all goes well this can go in -- we can work on documenting that map when the animation code gets hit by the refactoring sweep.

aashish24:
:+1: in general we should document our code as much as possible atleast when something is not trivial or  error prone. In my document, I will make it a requirement.

dlonie:
LGTM, tests pass.

Packages/vcs/Lib/VTKPlots.py
doutriaux1: yes but aren't we changing all of this anyway once we refactor?

PR #1240, opened 2015-04-29
Fixes typo in CDMS raise statements

remram44:
See [this ctest output](https://open.cdash.org/testDetails.php?test=331767198&build=3792026)

### DIFF
--- a/Packages/cdms2/Lib/dataset.py
+++ b/Packages/cdms2/Lib/dataset.py
@@ -84,26 +84,26 @@ def setCompressionWarnings(value=None):
         value = not _showCompressWarnings
     if isinstance(value,str):
         if not value.slower() in ['y','n','yes','no']:
-            raise CDMSError,"setCompressionWarnings flags must be yes/no or 1/0, or None to invert it"
+            raise CDMSError("setCompressionWarnings flags must be yes/no or 1/0, or None to invert it")
         if value.lower()[0]=='y':
             value = 1
         else:
             value = 0
     if not isinstance(value, (int,bool)):
-        raise CMDSError, "setCompressionWarnings flags must be yes/no or 1/0, or None to invert it"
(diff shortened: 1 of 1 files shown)


doutriaux1:
@remram44 I will check to see if the server is even up at all. Thanks.

aashish24:
@remram44 could you please make a request against release? Master will need to move on. please refer to my email earlier.

Packages/cdms2/Lib/dataset.py
```diff
@@ -245,14 +245,14 @@ def openDataset(uri,mode='r',template=None,dods=1,dpath=None, hostObj=None):
...
-                raise "Error in DODS open of: ",uri
+                raise CDMSError, "Error in DODS open of: ",uri
         else:
             try:
                 datanode = loadURI(uri)
                 return datanode
             except:
                 datanode = loadURI(uri)
-                raise "Error in loadURI of: ",uri
+                raise CDMSError, "Error in loadURI of: ",uri
```
doutriaux1: well if we really want to make this python proper shouldn't this be:
```
raise CDMSError("Error in loadURI of %s" % uri)
```

remram44: `raise ExcType, argument` is kind of an old way to write things, but is compatible with 2.x versions; I sticked with it since this is what is used throughout this file.

doutriaux1: I'd say let's make it clean so it will work when/if we move to Python 3.0. Since you brought that up, would you mind fixing it across the file? Thanks! Because the way it is now is also compatible with 2.x, so why bother changing it in the first place?

remram44: I'm certain 2to3 handles that, but sure, I can fire a regex at it.

doutriaux1:
awesome! Thanks @remram44

PR #1243, opened 2015-04-29
Fixes typo in CDMS raise statements

remram44:
#1240

Re-opened so that @aashish24 can merge without using Git.


aashish24:
Looks good to me@ but I will let @doutriaux1 have merge it. Thanks @remram44

doutriaux1:
@remram44 see my comments

remram44:
I went with the style already used in that file. I agree that it's an old syntax, but it's still recognized by Python 2.

PR #1252, opened 2015-05-01
Issue 1251 cdms2 file greater 2 gb

doutriaux1:
@aashish24 @chaosphere2112 @dlonie needed for ACME so they write BIG files


dlonie:
Change looks fine, but I think this would be a good exception for the "must test every change" rule :) Not that these exceptions should happen often, but this test has some practical issues:

- It writes out a file that is >3.0 GB is size (not sure how big it gets in the end, but when I checked it was at 3GB). This is an extreme amount of space to use during a unit test.
- What happens if there isn't sufficient disk space available? I didn't test this, but I suspect that the netcdf library will throw an exception, the call to `os.remove` won't execute, and future tests will fail as a result of not being able to allocate new files. This leaves the filesystem in a bad state and the user won't know what happened, why their filesystem is now full, or how to correct it without doing a lot of digging around.
- The test takes about 8 minutes to complete on my system. This is excessive, IMO.
- This feature is netcdf functionality, not UV-CDAT. Given the extreme cost of the test in size and time, I think it'd be ok to let netcdf's QA system test that this works, rather than us verifying that their code works as advertised.

Thoughts? The change itself looks good, but I'd vote to remove the test in this case.

doutriaux1:
@dlonie fine with me I really designed the test to make sure my changes worked. Taking it out right now.

dlonie:
Cool :+1: Can you remove the unused test script from the repo, too?

doutriaux1:
@dlonie it's done, removed the test.

doutriaux1:
thanks I'll take a look, maybe some non commited files.

Issue #1084, opened 2015-02-26
Error building pyOpenSSL on OSX


remram44:
With my version of OpenSSL, pyOpenSSL doesn't build:

```
OpenSSL/crypto/crl.c:6:23: error: static declaration of 'X509_REVOKED_dup'
follows non-static declaration

 static X509_REVOKED * X509_REVOKED_dup(X509_REVOKED *orig) {
                       ^

In file included from /opt/local/include/openssl/ssl.h:156:0,

                 from OpenSSL/crypto/x509.h:17,
                 from OpenSSL/crypto/crypto.h:30,
                 from OpenSSL/crypto/crl.c:3:

/opt/local/include/openssl/x509.h:751:15: note: previous declaration of
'X509_REVOKED_dup' was here

 X509_REVOKED *X509_REVOKED_dup(X509_REVOKED *rev);
               ^
```
Googling for this points me to [this post](http://forums.gentoo.org/viewtopic-t-1003484-view-previous.html?sid=f4549e0d47f947b304457dbfbb3ecbef), where it says that pyopenSSL 0.14 doesn't have the issue. UVCDAT uses 0.13 (from [uv-cdat.llnl.gov/cdat/resources/](http://uv-cdat.llnl.gov/cdat/resources/pyOpenSSL-0.13.tar.gz)).

durack1:
@remram44 it seems that Python 2.7.9 comes with SSL which has been backported from v3: https://www.python.org/downloads/release/python-279/

I wonder if the built-in SSL would work here?

remram44:
I tried replacing the version in `pyopenssl_pkg.cmake`, since 0.14 is on uv-cdat.llnl.gov, but that doesn't build either...

doutriaux1:
interstingly on my system:
```
doutriaux1@sofia:[~]:[3928]> cd build_many_j/build/PYOPENSSL/
/Users/doutriaux1/build_many_j/build/PYOPENSSL
ldoutriaux1@sofia:[PYOPENSSL]:[master]:[3929]> ll
total 768
-rw-r--r--   1 doutriaux1  staff   1.6K Feb 26 14:24 COPYRIGHT
-rw-r--r--   1 doutriaux1  staff   1.3K Feb 26 14:24 LICENSE
-rw-r--r--   1 doutriaux1  staff   6.1K Feb 26 14:24 Makefile.in
-rw-r--r--   1 doutriaux1  staff   4.4K Feb 26 14:24 README
[...]

PR #1253, opened 2015-05-01
Update PyOpenSSL to 0.14.

dlonie:
#1084 is popping up for me again on Arch Linux. This bumps PyOpenSSL to 0.14, which fixes the issue.

I'm targeting master for this, let me know if it should go into release instead/also.

@doutriaux1 @chaosphere2112 @aashish24


aashish24:
in theory, master should be good for new changes. I won't suggest it for release for now unless @doutriaux1 thinks otherwise.

dlonie:
Can we get this merged soon? I have to cherry-pick it every time I pull a branch at the moment.

aashish24:
Since we already have a release branch. This can go in without further delay. @doutriaux1 can you do a quick test on it? 

@dlonie since we have buildbot working now (not before), can you push a dummy commit on this branch so that we can trigger bulid again?

remram44:
I can't build this on my machine, I get weird C compiler errors... [PYOPENSSL-install-err.log](https://gist.githubusercontent.com/remram44/68be5be37ff83c88b474/raw/PYOPENSSL-install-err.log)

doutriaux1:
@remram44 so this is a 10.10 machine right? but the logs seem to indicate it's picking up 10.9 SDK. Just for now, can you please try to move your 10.9 SDK directory so that it's not picked up? 
@aashish24 we should really switch this all SDK logic so that it picks the SDK that matches your OS, not just the lower or greater one... I'll open an issue for this. @remram44 can you please tell me if moving the SDK fixes things for you for now. Thanks.

dlonie:
I had issues building this when I had the setup_runtime.py sourced, since the pkg-config in the UVCDAT build won't detect the system libraries properly. Dunno if that's related, but it might be a similar issue.

aashish24:
@doutriaux1 that will be hard for reasons that a user can have multiple SDK's. I think my other branch addresses that. If you have 10.10 SDK, it will pick that first since we know that you installed that and therefore you moved on to the new version.

doutriaux1:
@aashish24 why is it hard?
```
uname -a
Darwin sofia.llnl.gov 14.3.0 Darwin Kernel Version 14.3.0: Mon Mar 23 11:59:05 PDT 2015; root:xnu-2782.20.48~5/RELEASE_X86_64 x86_64
```
From that we know we should be using 10.10, no?

doutriaux1:
10.9:
```
Darwin maryam 13.4.0 Darwin Kernel Version 13.4.0: Wed Mar 18 16:20:14 PDT 2015; root:xnu-2422.115.14~1/RELEASE_X86_64 x86_64
```

aashish24:
its hard because a user can have 10.9 or 10.10 SDK both at the same time. What I was saying is that we can fall back to the latest SDK in that case. What OS you are on is not so relevant. I think whats more relevant is what XCode version you have with what SDK.

remram44:
Looks like Python didn't build correctly with the 10.10 SDK out of the way

doutriaux1:
and 10.8
```
uname -a
Darwin sofia.llnl.gov 12.6.0 Darwin Kernel Version 12.6.0: Wed Dec 17 19:11:40 PST 2014; root:xnu-2050.48.15~1/RELEASE_X86_64 x86_64
```

doutriaux1:
@aashish24 I disagree, the BEST guess is to match OS version and SDK. Then if you can't match them go ahead and do what we do.

doutriaux1:
@remram44 so you're on 10.9? Ok, starting a build on 10.9 to see if I can reproduce as well.

remram44:
I'm on 10.10. [This seems to be the issue](http://stackoverflow.com/questions/27976312/how-to-cope-with-non-gcc-compatible-code-in-os-x-yosemite-core-headers), compiling with clang now.

remram44:
With clang, it works with the 10.9 SDK out of the way, but not if I don't move it, as you had suspected.

doutriaux1:
ok. @aashish24 I will submit a branch where we match SDK to user's OS by default. Too many people seem to run into this issue.

doutriaux1:
ok new PR at: https://github.com/UV-CDAT/uvcdat/pull/1320

PR #1255, opened 2015-05-01
Marker deletion

chaosphere2112:
Fixes a bug with deleting the 0th marker in a marker object, adds a new test category (vcs/editors).

Passes all tests for me:
https://open.cdash.org/viewTest.php?buildid=3794900


aashish24:
:+1:  LGTM

PR #1259, opened 2015-05-01
Added new script to run tests

aashish24:


aashish24:
@doutriaux1 this will remove the requirement of souring the setup script just to run ctest. If you want in release, I can push this branch on top of release as well.

doutriaux1:
@aashish24 let me try it and then yes ;et's merge into release not master. BTW we really need to clean up the workflow, release vs master vs devel. What's the status with the document/draft?

aashish24:
@doutriaux1 the workflow of release is going to be as it is. We don't have a devel? What you are referring to. release and master workflow is there since last two releases. It is as follows: 

1. master is the default starting branch when no release in sight. 
2. Any time someone wants to fix something in release, they should start their work based off release
3. When we make a new release, we merge the last release into master using -s ours (we merge all other changes before that into master) 

That's it. From the user/developer point of view its only 1) and 2) are relevant. So its only 

release and master and no devel (I am not even sure why referred it here). I think may be you are referring to uvcmetrics?

I will have the document soon. I am working on integration framework which is almost close to done.

aashish24:
@doutriaux1 don't merge this one yet. I will let you know.

Issue #922, opened 2014-11-24
UV-CDAT code test coverage


durack1:
@doutriaux1 @aashish24 I think this might be something worth taking at look at http://blog.coveralls.io/2012/12/14/Coveralls-Open/

It seems to integrate neatly with Travis..

durack1:
@doutriaux1 another 2.3 candidate..

durack1:
@dnadeau4 just pinging you on this issue/query as test coverage might be something that may be useful within `cdms2` updates..

dnadeau4:
I'll take a look I think we really need something like this.

doutriaux1:
@chaosphere2112 to add to list of 3.0 new things

durack1:
@chaosphere2112 pinging you here.. Should this be a `2.6er`?

chaosphere2112:
@durack1 I actually have this almost entirely ready in a branch. I'll try and formalize it and submit a PR in the next couple of days.

durack1:
@chaosphere2112 even better! The versioneye service is pay for service.

chaosphere2112:
@durack1 Wrong issue :wink:

durack1:
@chaosphere2112 not really https://github.com/UV-CDAT/uvcdat/issues/922#issuecomment-64206531, but it should be - I'll purge that (versioneye) comment as it's been moved to a new dedicated one #1842

durack1:
@chaosphere2112 is this fixed, I think you did in #1851? If so wanna close?

aashish24:
@durack1 we can close this one. There is anohter issue I need to create is on submitting covergage to cdash but i will file another issue for it.

durack1:
@aashish24 I would have kept this open as the `cdash` reporting is part of a "final" solution I would have guessed? What's the point in creating another issue entry for the same thing?

aashish24:
> @aashish24 I would have kept this open as the cdash reporting is part of a "final" solution I would have guessed? What's the point in creating another issue entry for the same thing?

I think its better to have fine level resolution in general than try to have a issue that is long standing. Also, I think I would argue that we do have to the coverage now. Now having them reported to CDASH is a different problem.

durack1:
@aashish24 fair enough, but closing one issue and opening another is status quo in making progress against the 2.6 milestone..

aashish24:
> @aashish24 fair enough, but closing one issue and opening another is status quo in making progress against the 2.6 milestone..

[...]

PR #1290, opened 2015-05-06
Add pep8

doutriaux1:
@aashish24 @dlonie @remram44 @chaosphere2112 correct branch this time.


remram44:
I'm not sure there's a point in this, it is a development tool that the user will already have if he's using it. And activating the UV-CDAT environment probably doesn't interfere with it.

Sticking to a particular code style for UV-CDAT would probably be useful though.

doutriaux1:
but it doesn't hurt to have it. And for example our users here (and at other universities) only use UV-CDAT's Python so I'm trying to get them a complete set of tools. It's tiny too.

CMake/cdat_modules/pep8_pkg.cmake
```diff
@@ -0,0 +1,18 @@
...
+
+set (nm PEP8)
+string(TOUPPER ${nm} uc_nm)
+set(${uc_nm}_VERSION ${${nm}_MAJOR}.${${nm}_MINOR}.${${nm}_PATCH})
+set(PEP8_SOURCE ${PEP8_URL}/${PEP8_GZ})
+
+if (CDAT_BUILD_ALL)
+  add_cdat_package(pep8 "" "" ON)
+else()
+  add_cdat_package(pep8 "" "" OFF)
```
aashish24: it should be OFF even in ALL mode. ALL is meant for users and PEP is a developer tool. In near future, I would like to test ALL as well and more you will add to ALL, it will make it difficult to test it on dashboard machines. 

I like the idea and wanted to talk about it.

aashish24:
Nice to have it +1 other than my request it looks good.

CMake/cdat_modules/pep8_pkg.cmake
remram44: Since pip is available, users are a lot more likely to "pip install" it than remember to activate it before building, if it's not enabled by default. Just my two cents of course, I have no strong opinion about any of this.

aashish24:
Sure. I am thinking continuous testing / building point of view. Anytime we add a package we: 

1) Increase the changes that it will fail on some systems
2) Build time increases. 

May be should change the name to ALL to something else (SUPER) and then have ALL will build all of the packages. SUPER or something like that makes sense since we have. LEAN and  DEFAULT for now.

CMake/cdat_modules/pep8_pkg.cmake
doutriaux1: we can't use pip at the lab... firewall blocking us... @aashish24 ok I like to have it on only in full build.

doutriaux1:
@aashish24 let's think about it. It gets complicated we have LEAN/DEFAULT/ALL the more add the more confusing it gets.

dlonie:
I'm curious about the point of this, too. Just to have it available? I'm not sure developer tools really belong in the superbuild.

Now, if you wanted to use this (and ideally, a linter) to add tests that check the source code for consistency and correctness (and were willing to clean up the existing issues these tools would uncover), then I'd be 100% behind it...

I won't object to it going in, but if it does IMO it should be disabled by default for all builds, and only built when explicitly requested in cmake. The superbuild is already quite bloated.

dlonie:
Actually, if it goes in, make it on by default and I'll add a unit test that tests the new vcs.vcsvtk submodule. I'd love to have patches with poor readability rejected by a test for that submodule. Hopefully others would follow suite and do the same for their code.

aashish24:
@dlonie I liked the idea. For the testing and dashboards, we can setup the cmake to turn it ON. I would suggest we leave it OFF in all cases and tweak config for dashboards to turn it ON. What you all think? 

@doutriaux1 I agree 3 options are good. Adding 4th could be confusing. I was just proposing a solution but wasn't strongly favoring it.

dlonie:
How about it gets turned on anytime testing is enabled? I'd like any dev running tests to see issues before the pull request gets made.

durack1:
@dlonie @aashish24 @doutriaux1 I wonder if code coverage is something that should also be considered? #922 suggested this..

dlonie:
Improving test coverage would certainly improve the quality/reliability of the code. I'm not familiar with how coverage is tested in python, so I'm not sure how much effort it would take, though.

aashish24:
+100000000 for coverage. I was about to work on it. @jbeezley  has done some work. I will talk to him this afternoon. Yes, we need code coverage. Earlier, this week, I thought about making it as a first priority since only then we know exactly how much code we are testing.

chaosphere2112:
@dlonie @durack1 I've been working on getting some test coverage numbers with the tests I've been writing for the `vcs.vtk_ui` submodule. There's a module called [`coverage.py`](http://nedbatchelder.com/code/coverage/) that I've integrated with my tests (though all of my tests follow a convention, and derive from the same base class). I'm currently working on pushing my coverage up above 90%– sitting at 62% right now on a branch.

aashish24:
@dlonie How about it gets turned on anytime testing is enabled? I

+1. Do you want to push that change? I liked that idea. 

- Aashish

aashish24:
@chaosphere2112 yes that's what @jbeezley  uses as well: https://github.com/OpenGeoscience/gaia/blob/master/CMakeLists.txt

chaosphere2112:
@aashish24 @jbeezley [Here's the script](https://github.com/chaosphere2112/uvcdat/blob/4c26984bdaed95e8bedc7555db4917e6a71a788a/testing/vcs/vtk_ui/vtk_ui_coverage.py) I've been using to get some coverage numbers; at one point I modified it to get numbers for the bulk of the VCS tests, but it did involve some substantial editing of the existing tests, so I just did it as an exploration.

dlonie:
@aashish24 I won't have time to today, but I can look into that next week if no one beats me to it.

remram44:
The [coverage tool](http://nedbatchelder.com/code/coverage/) is great and pretty straightforward (it does slow down the code, though). Free service coveralls.io can display coverage status and alert/comment on PRs, works great with Travis ([see here](https://coveralls.io/r/remram44/usagestats?branch=master) for example)

chaosphere2112:
@remram44 Yup, that's what we're all looking at :+1:

dlonie:
Well, we don't need a tool for our travis coverage -- it's '0'.

Just noticed yesterday that it doesn't actually run any tests...

chaosphere2112:
@dlonie I think that was part of the shift over to buildbot; the tests were almost always failing on travis

durack1:
It seems you're spoilt for choice with this, https://github.com/z4r/python-coveralls and https://github.com/coagulant/coveralls-python are two other options listed off https://coveralls.zendesk.com/hc/en-us/articles/201342869-Python

jbeezley:
Coveralls works with the coverage.py output files.  It is [easy](https://github.com/OpenGeoscience/gaia/blob/master/.travis.yml#L42) to get working.  As for coverage.py, it is easiest to just run your tests with the coverage script.
```
python -m unittest ... ==> coverage -m unittest ...
```
or
```
python testscript.py ... ==> coverage testscript.py ...
```
The only difficulty is getting coverage to execute the correct python binary, use the correct python path, etc.

dlonie:
Sounds like the new buildbot `runtest` script could take care of calling the coverage tool in the correct environment for us.

chaosphere2112:
@dlonie That does sound easier than integrating it into the test process for every developer; would probably encourage people like me (who like to drop the massive PRs :wink:) to submit earlier, so we can get coverage numbers while we're in-progress.

aashish24:
right. the path should not be a problem for us anymore. Also, 

* we cannot run tests on travis ci because of time limitation and various other issues
* we moved to use buildbot for that purpose. travis is used only to test builds
* I can add code coverage stuff with help from @jbeezley and @chaosphere2112 
* @remram44 coverage.io looks good. I have see that before as I think you are using it for reprozip

remram44:
@aashish24 The ReproZip situation is a bit weird since I need coveralls to handle both C code and Python code coverage, which come from two different Travis jobs... It's in need of fixing unfortunately.

It did work at one point, though.

aashish24:
thanks @remram44.

aashish24:
@doutriaux1 this branch should go into master. I don't think its needed in release.

aashish24:
I am closing this pull request and will create a new one based on master.

PR #1299, opened 2015-05-07
Issue1212 durack1 spyder2.3.4 update rebased

durack1:
@doutriaux1 @aashish24 re-based off current master, this seems to have pulled everything across. Will check to see if the buildbot changes are triggered


jbeezley:
The buildbot won't trigger in this case because the change is coming from a forked repository.  In part, this is intentional because anybody can make a PR.  If we just blindly ran the tests, it would allow an attacker to execute arbitrary code on the builder.  I believe I can trigger a build manually after inspecting the PR.  I'll give that a try here to see what happens.

durack1:
@jbeezley maybe then I should back out the one file changes into a new PR then? The only file that I've updated is CMake/cdat_modules/spyder_pkg.cmake - it's literally just a tarball update

jbeezley:
That won't help unless you push the branch to this repository.  Someone should probably add you if you don't have ability to do that.

durack1:
@jbeezley to be honest, none of the tests were interacting with this change anyway, as it required a complete build to even touch the spyder sources.. If that's the case, then maybe it's not worth worrying about - unless of course you're testing the FULL builds in the buildbot tests?

jbeezley:
Okay, no problem.  I'm mostly just interested working out kinks in the buildbot.

durack1:
@jbeezley can I advocate that the buildbot tests run the FULL builds too? It would be great if even these peripheral changes to specialised packages were tested..

jbeezley:
I think that is the plan, we are just trying to get the issues with the default build worked out first.

durack1:
@jbeezley that's great, forward progress for sure!

durack1:
@doutriaux1 @aashish24 @jbeezley this PR should be deprecated by #1300 which seems to have triggered the buildbot testing

PR #1289, opened 2015-05-06
Add pep8

doutriaux1:
@aashish24 @dlonie @remram44 @chaosphere2112

PR #1303, opened 2015-05-09
Add pep8 master

aashish24:


aashish24:
#1289 #1290

CMake/cdat_modules/pep8_pkg.cmake
```diff
@@ -0,0 +1,16 @@
...
+set( PEP8_URL ${LLNL_URL} )
+set( PEP8_GZ pep8-${PEP8_VERSION}.tar.gz )
+set( PEP8_MD5 c7a3f57d832484a6295164661fbb1335 )
+
+set (nm PEP8)
+string(TOUPPER ${nm} uc_nm)
+set(${uc_nm}_VERSION ${${nm}_MAJOR}.${${nm}_MINOR}.${${nm}_PATCH})
+set(PEP8_SOURCE ${PEP8_URL}/${PEP8_GZ})
+
+if (BUILD_TESTING)
```
doutriaux1: do we turn on build testing? is it on by default?

aashish24: It is on by default.

aashish24: We can turn it OFF by default but I think its mostly us (developers) building it for now.

doutriaux1: if we turn it off, does that means ctest will not work? If so, leaving it ON is probably best.

aashish24: Right, there won't be any ctesting if it is OFF. I agree with you having testing ON by default is good. Having PEP installed when testing is ON is good IMO. If you agree, we can have it merged and then @dlonie can add style testing.

aashish24:
I am merging this one since this wasn't really my  branch and I approved it

Issue #1015, opened 2015-02-09
Unittests always reported as passed


jbeezley:
Many of the unittests in UV-CDAT are broken because ctest will always report passed even when it actually fails.  For example, see [here](https://open.cdash.org/testDetails.php?test=311441599&build=3687597).  This happens because the exit code is never set for these tests.  Either the tests need to use [`unittest.main`](https://docs.python.org/2/library/unittest.html#unittest.main), which will set the status automatically, or get the results via
[`TestResult.wasSucessful`](https://docs.python.org/2/library/unittest.html#unittest.TestResult.wasSuccessful).

dlonie:
Just making a note of this here, since it's related to this issue and doesn't have its own:

https://github.com/UV-CDAT/uvcdat/blob/be51764d9d410e114ee6011b59338946fad925ee/testing/dv3d/TestManager.py#L142

Missing baselines should be a failure.

jbeezley:
Bumping this again as I noticed most tests inside [`testing/regrid`](https://github.com/UV-CDAT/uvcdat/tree/master/testing/regrid) (~55 of 65) and [`testing/cdms2`](https://github.com/UV-CDAT/uvcdat/tree/master/testing/cdms2) (~20 of 45) don't set exit statuses when they fail.

PR #1307, opened 2015-05-11
Fixed the ordering of images for regression testing

aashish24:
So that the cdash will show them correctly in the right
order

@jbeezley @doutriaux1 @ThomasMaxwell  please review

### DIFF
--- a/testing/dv3d/TestManager.py
+++ b/testing/dv3d/TestManager.py
@@ -136,7 +136,7 @@ class vcsTest:
         try:
             print "Copying ref image %s to %s in %s " % ( self.image_name, ref_image, os.path.abspath('.') )
             shutil.copy( self.image_name, ref_image )
-            ret = checkimage.check_result_image( ref_image, test_image, checkimage.defaultThreshold+3. )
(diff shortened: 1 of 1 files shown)


doutriaux1:
Thanks!

jbeezley:
That caused the error to go away completely when it should have just switched the images in cdash.  This [line](https://github.com/UV-CDAT/uvcdat/blob/be51764d9d410e114ee6011b59338946fad925ee/testing/dv3d/TestManager.py#L142) makes me very suspicious.  I think an `IOError` is raised in the try block somewhere causing the test to return successfully when it should actually fail.  If there really is a reason to do this, the condition should probably be more narrowly defined.

doutriaux1:
what are you saying @jbeezley ? You don't like "magic"? 
More seriously should we worry about this now, or open a new issue for 2.3? I'll leave it up to you.

jbeezley:
I would say leave it for 2.3 and go over all of the tests to make sure failures are reported correctly (as in #1015).

aashish24:
@jbeezley agree with you. it should have ret = (anything but 0). I will push a branch post 2.2 release.

PR #1313, opened 2015-05-11
Issue 1311 new dap url

doutriaux1:
@aashish24 @jbeezley just for stability, since our esgf-cet sometimes goes down.
I think the error we (I) get is due to missing curl.h headers.


doutriaux1:
some file that shouldn't be here re-issuing

doutriaux1:
should be to release not master

PR #1342, opened 2015-05-22
Sync release

aashish24:
This is the only commit that's missing in master. @doutriaux1 @dlonie please approve.


dlonie:
There's a failure on buildbot, is that related to this? I'm guessing not since it's on a linux box....

The change LGTM, but I'll let someone with a mac test/merge it.

aashish24:
I am looking into that. One thing I am also working on is to sync repo names in testdata and the branch name as I would like to have dashboard 100% green.

durack1:
@aashish24 would it be possible to get some mac buildbots included here? Seems like there's a need for them..

aashish24:
@durack1 I totally agree with you. @doutriaux1 do we have a mac machine for this purpose? I am hoping that we will have the buildbot server out next week.

doutriaux1:
@aashish24 it's hard at the lab because of firewall issues and http access that needs to be MANUALLY renewed every 6 hours....

aashish24:
ahh...okay, I will see if I can find something here.

doutriaux1:
I will talk to Tony and see if we can move a mac to the green network, then we could setup a buildbot. In the past Tony couldn't do it easily, this might have changed.

Issue #1344, opened 2015-05-26
pep8 unit test


dlonie:
Add a test that runs `pep8` against the `vcsvtk` module. Make it easy to extend to other modules as well.

doutriaux1:
good idea! @dlonie can you also always add a milestone when you create an issue. It makes it easier to track via filters. Especially when we get close from a release. Thanks.

dlonie:
Sure, no problem.

dlonie:
@aashish24 Here's the pep8 test (#1351) -- it was merged already :+1:

PR #1351, opened 2015-05-28
Pep8 test

dlonie:
Addresses #1344.

@aashish24 @chaosphere2112 @doutriaux1


dlonie:
Note that this only enables style checking for `vcs/Lib/vtkvcs/*.py`.

aashish24:
Nice!!

PR #1363, opened 2015-06-04
Moves sample data to ${INSTALL}/share/

### LINKED ISSUE #1362: Sample data is all over the place
On Linux, sample_data is in `${CMAKE_INSTALL_PREFIX}/sample_data`; this is fine.

On Mac OS, sample_data ends up in `${CMAKE_INSTALL_PREFIX}/Library/Frameworks/Python.framework/Versions/Current/sample_data`.

Anywhere, the data for uvcmetrics ends up in `${HOME}/uvcmetrics_test_data`; this is inconvenient and wrong.

remram44:
Fixes #1362

Moves sample_data to share/uvcdat/sample_data
Moves uvcmetrics data to share/uvcmetrics/test_data

It seems to me that we might want `CDAT_DOWNLOAD_UVCMETRICS_TESTDATA` to default to off; it is over 1.3GB in size.


doutriaux1:
@remram44 did you run ctest on this? All vcs test look for the sample data in vcs.prefix+"/sample_data"

You will probably need to edit these as well.

remram44:
This is annoying because sys.prefix gets weird values on Mac OS. Will need to configure() a Python file for that.

doutriaux1:
that's why we use vcs.prefix (not sys.prefix in case you installed somewhere else than your python root)
Because that way it gets consistent between mac and not mac (su=ince sample_data is dumped currentlyat the same location so we don't need code for mac and non mac).  But that is because sample_data was not installed under share as this issue points out. Thanks for being willing to take the time to fix this.

aashish24:
@remram44 can you squash "Replace sample_data path" into single commit? It looks like they have the same commit message so seeing them in git history is kind of odd if that's Okay.

remram44:
They are different regexes... I'll squash once the tests pass.

aashish24:
Thanks, I understand, appreciate it.

remram44:
Somehow geos5-sample.nc doesn't end up in the right place (but still in the old install/sample_data/ location), making [7 tests fail](https://open.cdash.org/viewTest.php?onlyfailed&buildid=3844750).

aashish24:
I am going to push  a fix for it with a new updated branch and therefore closing it.

doutriaux1:
why closing BEFORE the work is done?

remram44:
@aashish24 Note that you can just update pull requests if the branch is from UV-CDAT/uvcdat

aashish24:
Thanks @remram44. The only reason I did because I appended _updated in the branch name and didn't want to loose the original branch just in case if I messed up something but i will remember next time.

PR #1369, opened 2015-06-04
1358 remove auto api

doutriaux1:
and vcs_legacy

should fix #1358 
@dlonie @sankhesh please review

cdash is at:
https://open.cdash.org/buildSummary.php?buildid=3844786

@dlonie wonder why pep8 test is skipped

PR #1390, opened 2015-06-14
Sample data location updated

aashish24:
Updated branch for #1363.


doutriaux1:
@aashish24 I would not stick the path into vcs.sample_data 
While most of the sample_Data are being used by vcs, some are also used by cdms2 and other packages. 
There is a "cdat_info" module (I know @remram44 it is created by the horrible install.py.in and I need to change this!) but that seems like the logical place for this sort of things.

doutriaux1:
I see you actually have such function in cdat_info. Perfect! Can you please use it accross the whole repo? So it is consistent.

remram44:
@doutriaux1 The path is `cdat_info.get_sampledata_path()`, I also added it to vcs as `vcs.sample_data` because it already had `vcs.prefix = cdat_info.get_prefix()`.

Changing `vcs.sample_data` to `cdat_info...` would require importing cdat_info everywhere.

doutriaux1:
also the changes in vcs_legacy are going to create a merge conflcit with PR #1369 which removes this directory all together.

doutriaux1:
@remram44 I guess it's ok, I don't really mind it, just thinking about a new user trying to create a new test and copying from these, they would never know about the "proper" way. I can leave with it though. Should I merge then (after I ran ctest on this of course).

aashish24:
@doutriaux1 that is a good point. I think it would be better down the road to have something other than vcs managing the data path but for now I think vcs is Okay. One think I should mention that importing vcs (before build) is not ideal since it seems that it needs vtk and other modules imported as well which won't be found until build.

PR #1403, opened 2015-06-22
Make tests fail if baselines not found

sankhesh:
Refrain from redundant image copies while testing.

@dlonie: Please review

### DIFF
--- a/testing/checkimage.py
+++ b/testing/checkimage.py
@@ -61,11 +61,6 @@ def check_result_image(fname, baselinefname, threshold, baseline = False, cleanu
     else:
         baselinefnames = [baselinefname,]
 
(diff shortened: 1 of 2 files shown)


testing/checkimage.py
```diff
@@ -61,6 +61,11 @@ def check_result_image(fname, baselinefname, threshold, baseline = False, cleanu
     else:
         baselinefnames = [baselinefname,]
 
+    for b in baselinefnames:
```
dlonie: Let's take this bit out -- The VTK image testing backend will pick up on the missing file and print an error while making sure the new baseline still gets uploaded to CDash.

dlonie:
One suggested change, otherwise looks good! :+1:

PR #1420, opened 2015-06-23
Add flake8 test for WK.

dlonie:
The test currently fails. The code needs to be cleaned up before merging.

@doutriaux1 @aashish24


doutriaux1:
@chaosphere2112 I forgot about this! We should add these to the gallery! I will see if I can move the tests to the test suite as well so that they show up in the gallery.

doutriaux1:
needs: https://github.com/UV-CDAT/uvcdat-testdata/pull/57

doutriaux1:
@aashish24 @dlonie I suspect the seg fault is related to: #1446 it seg fault for me when in bg. Let's wait on this PR then.

Issue #1400, opened 2015-06-22
vcs.Canvas.setantialiasing unusable


dlonie:
`vtkRenderWindow` needs to have multisampling settings specified before the first render, but currently `setantialiasing` can only be used before the window is created (which triggers an error) or after initialization (which has no effect).

The easy fix would be to make an antialiasing ivar in the canvas or backend that can be set between calls to `vcs.init` and `Canvas.plot`, and warns if called after the window is already initialized.

aashish24:
@dlonie I would assign it to @sankhesh

dlonie:
@sankhesh Grab me if you want to go over this -- this one should be higher priority than the other issues.

dlonie:
@sankhesh The code of interest is in `Packages/vcs/Lib/VTKPlots.py`, the `setantialiasing` and `getantialiasing` methods.

To visually determine whether or not antialiasing is working the following command line can be used to produce a line plot with aliasing artifacts:

`[build dir]/install/bin/runtest [source dir]/testing/vcs/test_vcs_basic_gms.py --gm_type=yxvsx --show`

sankhesh:
Fixed and merged

Issue #1079, opened 2015-02-25
Several marker problems/differences in 2.1.0 (compared to 1.5.1)


jypeter:
I have used test script listed at the end of this issue to test markers in 2.1.0 and see how they differ from 1.5.1. I have found several problems and differences, but @chaosphere2112 can probably use this script in the gallery once the problems are solved

I originally wanted to experiment with projections to see if #672 had been solved but I have found other problems and probably won't have time to experiment with non default projection. **My ultimate goal and need is of course to be able to overlay data points markers over a vcs map, whatever the projection is!**

 png created with 2.1.0
![test_simple_marker-2 1 0](https://cloud.githubusercontent.com/assets/2785573/6372426/69c2ca5c-bd04-11e4-96a7-3b567df63486.png)

png created with 1.5.1
![test_simple_marker-1 5 1](https://cloud.githubusercontent.com/assets/2785573/6372430/7ac27fbe-bd04-11e4-8e94-940d7f54c86a.png)

The problems and differences:
- The markers are smaller in 2.1.0 and 1.5.1 . You probably want to **check the scale to make sure to get the same size**, but _being able to get smaller markers than 1.5.1 is a good thing!_ So I hope we can use non integer marker sizes to be able to get lots of different marker sizes
- At least **one marker from 1.5.1 is missing from 2.1.0** and I get the following warning. You can also see the result on the plots above

```
/home/share/unix_files/cdat/versions/cdat_install_uv-2.1.0_x86_64_gcc4_13/lib/python2.7/site-packages/vcs/vcs2vtk.py:1113: UserWarning: unknown marker type: star, using dot
  warnings.warn("unknown marker type: %s, using dot" % t)
```
- **Non filled markers in 1.5.1 are filled in 2.1.0**! See the plots above: eg _triangle_up_ is the same as _triangle_up_fill_
- There is a _nice improvement_ I should mention! **The markers are much nicer in the 2.1.0 pdf than in 1.5.1**! But there is room for improvement. As you can see in the 300% magnified pdf screenshot below, with 1.5.1 on the left and 2.1.0 on the right, the _dot_ and _circle_ markers are much nicer in 2.1.0, but they should be described/rendered with a circle primitive (filled or empty) rather than with a list of segments
  ![magnified_pdf_vcs_markers](https://cloud.githubusercontent.com/assets/2785573/6372998/14f8cdec-bd08-11e4-9a89-90939fca6ece.jpg)
[...]

PR #1422, opened 2015-06-23
Make sure antialiasing is set right during context creation

sankhesh:
Fixes issue #1400

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -52,6 +52,8 @@ class VTKVCSBackend(object):
     self.numberOfPlotCalls = 0
     self.renderWindowSize=None
     self.clickRenderer = None
+    # Turn off anti-aliasing by default
+    self.antiAliasing = 0
 
     if renWin is not None:
       self.renWin = renWin
@@ -293,7 +295,7 @@ class VTKVCSBackend(object):
       self.renWin.SetStencilCapable(1)
       ## turning off antialiasing by default
       ## mostly so that pngs are same accross platforms
-      self.renWin.SetMultiSamples(0)
+      self.renWin.SetMultiSamples(self.antiAliasing)
       self.initialSize()
 
     if self.renderer == None:
@@ -926,16 +928,12 @@ class VTKVCSBackend(object):
       warnings.warn("Please implement gettextextent for VTK Backend")
 
   def getantialiasing(self):
-    if self.renWin is None:
-      return 0
-    else:
-      return self.renWin.GetMultiSamples()
+     return self.antiAliasing
 
   def setantialiasing(self,antialiasing):
-    if self.renWin is None:
-      warnings.warn("no RenderWindow ready, skipping setantialiasing call, please reissue at a later time")
-    else:
-      self.renWin.SetMultiSamples(antialiasing)
+    if self.renWin is not None:
+        warnings.warn("RenderWindow already created. setantialiasing call would have no effect here.")
+    self.antiAliasing = antialiasing
 
   def createLogo(self):
     if self.canvas.drawLogo:
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -53,7 +53,7 @@ class VTKVCSBackend(object):
     self.renderWindowSize=None
     self.clickRenderer = None
     # Turn off anti-aliasing by default
-    self.antiAliasing = 0
+    self.antialiasing = 0
 
     if renWin is not None:
       self.renWin = renWin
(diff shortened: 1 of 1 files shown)


sankhesh:
@dlonie Please review

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -293,7 +295,7 @@ def createRenWin(self,*args,**kargs):
       self.renWin.SetStencilCapable(1)
       ## turning off antialiasing by default
       ## mostly so that pngs are same accross platforms
```
dlonie: Let's remove this comment, since it's no longer true after this patch.

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -52,6 +52,8 @@ def __init__(self,canvas,renWin=None, debug=False,bg=None):
     self.numberOfPlotCalls = 0
     self.renderWindowSize=None
     self.clickRenderer = None
+    # Turn off anti-aliasing by default
+    self.antiAliasing = 0
```
dlonie: Let's just call it `self.antialiasing`, all lowercase (to match the set/get calls)

dlonie:
A couple minor issues, but looks good to me otherwise :+1:

sankhesh:
Implemented @dlonie's suggestions in e1b093d

aashish24:
Nice!! this will make things look much better. We should send an email once this branch is merged with before and after (pick a crappy looking plot from before)

dlonie:
Works here! @doutriaux1 Want to test/merge as well? @sankhesh and I are both at Kitware.

dlonie:
@aashish24 Here are the images I sent you and Brian while investigating this. At full resolultion it's quite an improvement.

**No multisampling (current master)**
orig

**16x MSAA (possible after this patch)**
with_msaa

**16x MSAA + `gm.linewidth = 2` (best match to Brian's sample plots)**
with_msaa_linewidth2

aashish24:
Nice! is 16x MSAA is set by you on your driver or someplace else?

dlonie:
The argument to `Canvas.setantialiasing` (and eventually `vtkRenderWindow::SetMultiSamples`) is the number of samples to take.

So after this patch, users can do:

```
x = vcs.init()
x.setantialiasing(16) # or 8, or 4, or 2, or 0...
x.plot(...)
```

and the result will be antialiased. If the graphics hardware/driver doesn't support the number of requested samples, VTK will set it as high as possible for the hardware/driver.

aashish24:
Okay, sounds good to me. Thanks. @doutriaux1, some folks want this today as they producing 1D plots. Would it be possible for you to give it a thumbs up?

jypeter:
@aashish24, can you try running the script in #1079 and see if you get nice(r) dots and markers, when using the smoothing?

doutriaux1:
great! building now and will merge.

aashish24:
@jypeter I will request @sankhesh to do it. @sankhesh can you post your results here?

doutriaux1:
@sankhesh k it works and it's great. Unfortunately for you it's way too great to be left to the user. So can I please ask you to tweak this so that antialiasing is on by default. Of course that means turning it off for all vcs tests... sorry.

aashish24:
@doutriaux1 we cannot turn it ON for testing since it will make resulting image vary depending on the user's setting. I would suggest we merge this one in and have @sankhesh push another branch where we turn it OFF for testing and having it ON by default.

aashish24:
Ooops..sorry I just read your email wrong. I totally agree with you and that's what we were planning to do it (as we do the same for VTK).

doutriaux1:
@aashish24 should we open another issue for this then? I was hoping @sankhesh would stick it in this branch.

doutriaux1:
done, see: https://github.com/UV-CDAT/uvcdat/issues/1433

sankhesh:
@jypeter Here is a montage of screenshots generated using the script in #1079 using different antialiasing modes:
screenshot from 2015-06-29 11 41 58

No antialiasing:
test_simple_marker_no_msaa

8x MSAA:
test_simple_marker_8x_msaa

16x MSAA:
test_simple_marker

jypeter:
Thanks @sankhesh , it does look nicer with 8x and 16x (8x may be enough)!

I don't remember if I have already asked this, but is it possible to generate nicer circles and even triangles from the start (before the smoothing stage)? When you look at the triangle in the "no MSAA" picture above, it looks pixelized, but with much bigger pixels than the pixels in the "trian" string

doutriaux1:
@sankhesh it's probably just a matter of tweaking the vtksource object we're using.

aashish24:
@doutriaux1 yes, that should be possible. I will create a new issue for it.

PR #1430, opened 2015-06-25
Issue 1425 bad flake8

doutriaux1:


doutriaux1:
@aashish24 @sankhesh @dlonie

dlonie:
In the future it's good to rebase things like https://github.com/UV-CDAT/uvcdat/commit/e242d56df7b1822bc2c0bfdd343b694424508bab into the patches they're correcting so that each commit compiles on its own. One day, it'd be great to have the option to use `git bisect` to track down UV-CDAT regressions, but we can't do that as long as there are broken commits in the history.

I can't really test this since it works around a LLNL-specific issue, but LGTM and the buildbots pass :+1:

doutriaux1:
@aashish24 I no speak english I sorry I am :smile:

Issue #1397, opened 2015-06-16
creating pngs in background mode leaks memory


doutriaux1:
A LOT!!!
```python
for i in range(120):
   x.plot(data,"default","isofill")
   x.png("file_%i.png" % i)
   x.clear()
```
run and look at memory in top.

doutriaux1:
@dlonie any idea why the memory could grow when bg=True ? The renWin is different obviously. Could it be VTK related?
Note: I tried to comment out hideGui/showGUI in def png but no luck

dlonie:
I'm a bit confused, it doesn't look like the `plot` call in the sample code is using background mode?

In any case, it doesn't look like VTK is holding onto anything in offscreen mode:

```
import gc
gc.enable()
gc.set_debug(gc.DEBUG_LEAK)

def runTest():
    import vtk
    sphere = vtk.vtkSphereSource()

    mapper = vtk.vtkPolyDataMapper()
    mapper.SetInputConnection(sphere.GetOutputPort())

    actor = vtk.vtkActor()
    actor.SetMapper(mapper)

    ren = vtk.vtkRenderer()
    ren.AddActor(actor)

    renWin = vtk.vtkRenderWindow()
    renWin.AddRenderer(ren)

    renWin.OffScreenRenderingOn()
    renWin.SetSize(300, 300)

    renWin.Render()

runTest()
del runTest

gc.collect()

leaks = {}

for i,g in enumerate(gc.garbage):
    t = type(g)
    print "Leak %d: %s @0x%x:\n   %s"%(i, t, id(g), str(g)[:70])
    leaks[t] = leaks.get(t, 0) + 1

for t in leaks:
    print "%7d: %s"%(leaks[t], t)

print len(gc.garbage),"leaks total."
```

```
$ python vtk_offscreen.py 
0 leaks total.
```

Looks like the leaks are on the VCS side of things.

doutriaux1:
@dlonie  I'm not quite sure what's going. All @durack1 and I noticed is that if you plot with vcs bg=1 then the process mem size keeps growing. And looking at the code it seems that the only differenc eis this:
```python
    if self.bg:
        self.renWin.SetOffScreenRendering(True)
        self.renWin.SetSize(self.canvas.bgX,self.canvas.bgY)
```

But obviously there must be something else.

durack1:
[...]

PR #1431, opened 2015-06-26
added a test for @aashish24

doutriaux1:
@aashish24 this adds a test that shows the leak.
if you turn bg=0 then you see less of a leak.
do not merge now, there's nothing to it.

### DIFF
--- a/testing/vcs/test_png_mem_leak.py
+++ b/testing/vcs/test_png_mem_leak.py
@@ -1,0 +1,24 @@
+import vcs
+import numpy
+import MV2
+import resource
+
+data = numpy.sin(numpy.arange(360*180))
+data.shape=(180,360)
+
+x=vcs.init()
+
+mem_previous = 0
+for i in range(10):
+    x.plot(data+i,"default","isofill",bg=True)
+    x.png(__file__+".png")
+    x.clear()
+    mem = float(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)/1024.
+    memStr = 'Iter: %i: Max mem: %.3g MB' % (i,mem)
+    print memStr,
+    if i>0:
+        memStr = ' Mem increase: %.3g MB' % (mem-mem_previous)
+    else:
+        memStr = ''
+    print memStr
+    mem_previous = mem


aashish24:
thanks @doutriaux1 do you see any VTK objects in the leak?

doutriaux1:
no idea, we should run it through the runTest but not in this branch because the dv3d will just drawn them in the noise.

aashish24:
Okay thanks. @dlonie how did you disable dv3d?

dlonie:
Did you mean @doutriaux1?

aashish24:
I meant @dlonie -:) When you were doing test on memory leaks, you were able to bypass dv3d module loading.

dlonie:
IIRC, i just removed the import statements that imported either dv3d.py from the VCS module or the DV3D module (I don't think there were any of the latter, though).

Be warned that I'm still seeing a ton of circular references and bound methods leaking when anything more involved than `import vcs` is run, so there will still be a lot of noise, even without `dv3d`.

doutriaux1:
@aashish24 this is somewhat urgent see: https://github.com/UV-CDAT/uvcdat/issues/1397

doutriaux1:
@sankhesh @dlonie I think tis is different from the leaks @dlonie wants me to look at. Can one/both of you please take a look. Thanks,

dlonie:
@doutriaux1 If you can reproduce this using pure VTK, I'll take a look, but working with VCS is beyond the current scope of my involvement in UV-CDAT.

Again, I'd suggest using tools like `objgraph` that let you track memory state and see what object are being created and not freed in between calls to `plot`. Given the number of memory issues we've had with VCS and that I've never heard of this happening with VTK before, I doubt the issue is on our end unless I see a script that reproduces this without VCS's additional layers clouding the waters.

doutriaux1:
@aashish24 I will let you find someone at Kitware to take a look at this.

aashish24:
@doutriaux1 did you try reproducing it in pure VTK?

durack1:
@aashish24 it would be great to solve this issue for `2.4.0` any chance of that? It's a pretty nasty usability bug that keeps tripping me over..

aashish24:
@durack1 look at the other thread on memory leak. Atleast we have a fix on vtk offscreen memory leaks now. We still have other memory leaks but I am hoping that we can start tracing those in near future as well.

PR #1437, opened 2015-06-30
R and rpy2

sankhesh:
Enable R and RPY2 support.

@doutriaux1 Kindly review

### DIFF
--- a/CMake/cdat_modules/r_external.cmake
+++ b/CMake/cdat_modules/r_external.cmake
@@ -8,6 +8,11 @@ else ()
     set(WITHX "yes")
 endif()
 
+list(APPEND USR_ENVS
+  "CPPFLAGS=-I${cdat_EXTERNALS}/include $ENV{CPPFLAGS}"
+  "LDFLAGS=-L${cdat_EXTERNALS}/lib"
+  )
+
 ExternalProject_Add(R
   DOWNLOAD_DIR ${CDAT_PACKAGE_CACHE_DIR}
   SOURCE_DIR ${R_source}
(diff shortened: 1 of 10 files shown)


sankhesh:
@aashish24

CMake/cdat_modules/r_external.cmake
```diff
@@ -16,7 +21,8 @@ ExternalProject_Add(R
   URL_MD5 ${R_MD5}
   BUILD_IN_SOURCE 1
   PATCH_COMMAND ""
-  CONFIGURE_COMMAND <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> LIBnn=lib --without-jpeglib --disable-R-framework --enable-R-shlib --disable-openmp --without-cairo --without-ICU --without-libpng --without-system-xz --without-aqua --without-tcltk --without-readline --with-x=${WITHX}
+  DEPENDS ${R_deps}
+  CONFIGURE_COMMAND env ${USR_ENVS} <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> LIBnn=lib --without-jpeglib --disable-R-framework --enable-R-shlib --disable-openmp --without-cairo --without-ICU --without-libpng --without-system-xz --without-aqua --without-tcltk --with-x=${WITHX}
```
doutriaux1: @sankhesh could you try to build uvcdat with -DCDAT_BUILD_PARALLEL=ON and turn on openmp in here?

Thanks.

CMake/cdat_modules/r_pkg.cmake
```diff
@@ -1,9 +1,9 @@
 set(R_MAJOR_SRC 3)
-set(R_MINOR_SRC 1)
-set(R_PATCH_SRC 0)
-set(R_URL ${LLNL_URL})
+set(R_MINOR_SRC 2)
+set(R_PATCH_SRC 1)
+set(R_URL "http://cran.r-project.org/src/base/R-3")
```
doutriaux1: ok I will tweak this back to LLNL when it works, it's "safer" to leave the links pointing to our server, less likely to break in the future.

CMake/cdat_modules/rpy2_external.cmake
```diff
@@ -0,0 +1,11 @@
+# create an external project to install RPY2,
+# and configure and build it
+set(nm RPY2)
+
+# Set PATH and R_HOME to find R
+list(APPEND USR_ENVS
+  "R_HOME=${cdat_EXTERNALS}/lib/R"
```
doutriaux1: @aashish24 when we enable the options to use system's externals, let's make sure to set this correctly.
@sankhesh could you set this only if 
1- CDAT_BUILD_R is ON and R_HOME is not already set?

CMake/cdat_modules/r_external.cmake
doutriaux1: @sankhesh we should check if user has R_HOME set at this point, is so we should error exit and tell user to turn either unset it or not build R from here.

aashish24: @doutriaux1 let's do then when we work on the system package overhaul.

CMake/cdat_modules/rpy2_external.cmake
aashish24: @doutriaux1 CDAT_BUILD_R will be ON if rpy2 is building since we are not supporting system R.

doutriaux1: ok let's wait until we allow for systemr to address this.

doutriaux1:
testring on my mac and ubuntu and merging.

doutriaux1:
should we add a basic test "import e2py" to make sure the build went properly?

sankhesh:
@doutriaux1 Yes. I'll be adding one soon.

sankhesh:
Replaced with #1440

PR #1440, opened 2015-07-03
R and rpy2

sankhesh:
Enable R and RPY2 support


sankhesh:
Taken over from #1437 

@aashish24 @doutriaux1  Please review

CMake/cdat_modules/singledispatch_pkg.cmake
```diff
@@ -0,0 +1,16 @@
+set( SINGLEDISPATCH_MAJOR 3 )
```
aashish24: why do we need this package?

aashish24: Can we not add this to the list of requirements?

doutriaux1: ubuntu needs it, and it's python based so can't be in requirements

aashish24: @doutriaux1 can you explain bit more on why Ubuntu needs it?

doutriaux1: setup.py on ubuntu lists it as a dependeny. So inside our firewall it died because it couldn't fetch it. It's yet another example of the "cool/sneaky" features of pip...

aashish24: Coo! thanks for the information. Very useful. 

- Aashish

PR #1452, opened 2015-07-10
Fix CONTRIBUTING.md

remram44:
It looks like this was merged without proof-reading or spell-checking.


aashish24:
Thanks @remram44!

Issue #1338, opened 2015-05-20
isoline labels don't work in overlay


potter2:
I think I submitted this before but the isoline labels don't appear when an isoline plot is an overlay on a isofill  plot

doutriaux1:
@remram44 assigning this to you guys because it works from the command line. It must have something to do with the order in which vistrails runs this.

![isolines_on_top](https://cloud.githubusercontent.com/assets/2781425/7738209/d64b1a82-ff06-11e4-8f8f-ea76f028300f.png)

remram44:
Running the generated script from cdat, I don't see isolines either.
```python
import cdms2, cdutil, genutil, vcs

cdmsfile = cdms2.open('/home/remram/uvcdat-builds/build-2.2.0-gd124bd7/install/sample_data/clt.nc')
clt = cdmsfile('clt')
clt = clt(latitude=(-90.0, 90.0),squeeze=1,longitude=(-180.0, 175.0),time=('1979-1-1 0:0:0.0', '1988-12-1 0:0:0.0'),)

canvas = vcs.init()
gmIsoline = vcs.getisoline('default')
canvas.plot(clt, gmIsoline, cdmsfile=cdmsfile.id)
canvas.interact()
```

doutriaux1:
well, yes, I don't know what this "cdmsfile=cdmsfile.id" is doing here. vcs probably error exit under the hood choking on this. So that's the issue.

doutriaux1:
Just to be clear, the scritp/VisTrails should NOT add this argument. I don't even know what this is supposed to do... @ThomasMaxwell is it something for DV3D?

doutriaux1:
@remram44 actually I'm not sure what you are talking about here, I do SEE isolines on the picture you posted...

remram44:
Yes but I don't see any label.

doutriaux1:
you need to turn them on manually. The default is no labels. I guess it's historical, users didn't want to see them on all the time.

remram44:
So... what's the issue?

doutriaux1:
@remram44 @potter2 I get it now... a bit slow... sorry... @potter2 are you saying that you did turn on the labels and they didn't show up?

doutriaux1:
@potter2 I thought you meant if you plot isoline on top of isofill then you don't see the isolines.

doutriaux1:
[...]

PR #1418, opened 2015-06-23
Add flake8 test for vcs.

dlonie:
The test currently fails. The code needs to be cleaned up before merging.

@doutriaux1 @aashish24

PR #1459, opened 2015-07-20
Create a new renderer for each new plot type

sankhesh:
Currently, the test whether the renderer exists is not sufficient
to decide whether a new renderer needs to be created. This was
causing different plots to use the same renderer.

This change introduces a new argument to the fitToViewport call
that forces it to create a new renderer if requested. This will
enable all the pipeline plot calls to create their own
independent renderer.

Fixes issue #1338


sankhesh:
@doutriaux1 @aashish24 @dlonie Please review

doutriaux1:
@sankhesh considering you had to change 6 files with the same edit, shouldn't fit2viewport be moved to piepline2d? (or even pipeline?)

doutriaux1:
@sankhesh will apporve this as soon as @aashish24 approved https://github.com/UV-CDAT/uvcdat/pull/1418 otherwise merge conflicts might arise and that will delay @aashish24 reviewing of it even further.

sankhesh:
There are some other places that fitToViewport is called. I'm not sure if all of that can be moved. Plus, we might be able to get rid of fitToViewport once the Context2D work goes in.

sankhesh:
Baseline for new test added in UV-CDAT/uvcdat-testdata@ca1be96

sankhesh:
@doutriaux1 

This change causes one test failure on my linux machine. It is the `text_vcs_basic_boxfill_masked_aeqd_proj` test.

With my changes I get the following output image:
test_vcs_basic_boxfill_masked_aeqd_proj
 

Whereas the test image looks like:
test_vcs_basic_boxfill_masked_aeqd_proj

Basically, now more area of the plot is visible and un-masked. Since, I haven't touched any masking code, I am inclined to think that the original baseline was wrong and these changes produce the right image. Do you concur? If yes, I will update the baseline.

doutriaux1:
@sankhesh this plot is worng in both version anyway.... Look at Africa it should have purple blob over sahara but it only has 1/2 the blob. This is another wrap issue with projections that we should look into. Let's update the baselines and record an issue for this "bad" plot.

sankhesh:
Thanks for quick reply @doutriaux1. Will do.

sankhesh:
@doutriaux1 

Projection issue documented here: #1462

aashish24:
@sankhesh can you update this branch so that we can merge it soon? Thanks.

sankhesh:
@aashish24 Fixed merge conflicts

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1033,7 +1037,8 @@ def scaleLogo(self):
                 self.setLayer(self.logoRenderer, 1)
                 self.renWin.AddRenderer(self.logoRenderer)
 
-    def fitToViewport(self, Actor, vp, wc=None, geo=None, priority=None):
+    def fitToViewport(self, Actor, vp, wc=None, geo=None, priority=None,
```
aashish24: the style seems off or may be the github is not showing me the right diff?

sankhesh: What do you think is wrong here?

sankhesh:
Continued in #1468

PR #1458, opened 2015-07-16
Enable Matplotlib in default build

sankhesh:
This addresses issue #1406

Issue #1406, opened 2015-06-23
Enable Matplotlib in default build


aashish24:

jypeter:
I completely agree with this! These days, most python classes teach new users how to use matplotlib for plots, so it will be hard to convince potential users to switch to the uv-cdat distribution if matplotib is not available by default... You may want to bundle basemap as well

And you also want to **convince new users that cdms2+vcs is easier to use than netcdf4+matplotlib+basemap**. Well, at least you need less lines in cdms2+vcs to get a basic plot, which is often enough for interns... @chaosphere2112 , maybe you can use a variation of http://uvcdat.llnl.gov/display.php?file=test_vcs_patterns on the web site to show people how easy it is to make a basic plot. At least for data on regular grids

 - @chaosphere2112 btw, why is the example from the gallery above named *patterns*? I don't see any patterns

 - @doutriaux1 can the *new* vcs handle again projections created with *createprojection* ?

aashish24:
@jypeter I personally think cdms2+vcs is way easier than netcdf4+matplotlib+basemap in terms of APi and ease of use. Also, unless you are doing mostly static plots (which I don't) matplotlib does not help as much. But I agree that we need to keep working on documentation, improving the code quality and stability, and improving the features to make it more convincing. Matplotlib is easy to deploy but basemap is not. We are trying to enable both by default.

jypeter:
I don't remember having problems installing basemap, I thik it worked fine when I used  ```-D CDAT_BUILD_BASEMAP=ON  -D CDAT_BUILD_MATPLOTLIB=ON```. It's cartopy that is hard to install

 - cdms2+vcs gives you default maps faster than other packages for data on **regular grids**, and that's something we can sell to new users
 - But there are more and more *people using basemap to create plots for data on irregular grids*, and basemap may work in cases where meshfill does not. I think the basemap graphics methods are not as picky as meshfill and do not require the specification of cell boundaries. But I'm no specialist and really suck at plotting this kind of data. Basemap also provides more graphics methods for irregular grids than just 'fill'. It may be interesting to see how basemap can actually plot things without knowing about the cell boundaries, and how accurate the plots are
[...]

PR #1461, opened 2015-07-21
Basemap build

sankhesh:
#1406 #1458


sankhesh:
@aashish24 @doutriaux1 
Please review

doutriaux1:
@sankhesh can you please look into the pyside/qt5 error? We are using Qt4.

doutriaux1:
@sankhesh still tries to grap PySide

doutriaux1:
@chaosphere2112 where are we with converstion to PySide vs PyQt? Might not be worth fixing this if you are close.

chaosphere2112:
@doutriaux1 A ways out still; hoping to have a barebones version in the next month and half, and flesh it out from there.

sankhesh:
@doutriaux1 

I don't think this issue is related to qt.. I spoke to Aashish and for now, I am going to change the matplotlib backend to the default one instead of PySide/PyQt.

doutriaux1:
@sankhesh what is the "default" one?

sankhesh:
[Anti Grain Geometry (Agg)] (http://matplotlib.org/faq/usage_faq.html#what-is-a-backend)

doutriaux1:
Seems like it only does backgrond-like mode. PNG only. We need some sort of viz on screen directly.

It used to work perfectly what changed?

C.

sankhesh:
@doutriaux1 

The issue crops up only on the buildbot machine. Buildbot turns off CDAT_BUILD_GUI that in turn disables building of PyQt and SIP. Since matplotlib expects PyQt and doesn't find it, it looks for PySide. But that fails as well and matplotlib only reports a partial error. I'll try adding some logic to the matplotlibrc to switch to Agg backend if CDAT_BUILD_GUI is OFF.

doutriaux1:
Oh ok! Then it totally makes sense. Ok can you please tweak the build then to use Agg if GUI is OFF?

Thanks.

sankhesh:
@doutriaux1 

This branch is ready to be merged. The test failure is because the testdata was updated in a different branch. It should be fixed when this branch is merged.

doutriaux1:
passes! Yeah! https://open.cdash.org/viewTest.php?onlyfailed&buildid=3937013

PR #1468, opened 2015-07-24
1338 isoline labels

sankhesh:
Takes over from #1459


sankhesh:
@aashish24 

Please review

aashish24:
Thanks @sankhesh!

aashish24:
@sankhesh there should be a data branch as well I guess that we need to merge?

aashish24:
Eventually we need to get rid of fitTOViewport and I hope that we can have classes for each plot; each with its own Renderer. But for now it looks good to me.

aashish24:
+2 :+1:

PR #1490, opened 2015-08-13
Ansible vagrant install

zachmullen:


jbeezley:
This doesn't trigger other builders because it is from an external repository.  It doesn't really matter in this case because it is only devops content.

zachmullen:
I am not sure if it should even go in the main uvcdat repo or if I should just keep it on my fork, just wanted to put it out there for comments just in case.

jbeezley:
It's a little bit contrived for your use case, but so is the Dockerfile for cdatweb.  I think it is useful enough to serve as an example of a deployment method.  I'll wait for some else to give the final okay though.

doutriaux1:
@aashish24 I'm ok with merging this, I will let you have the final say.

aashish24:
@zachmullen @doutriaux1 I only have one suggestion. Can we move these files to its own specific folder (may be create deploy at the root level?) and then move docker and these there?

zachmullen:
That is possible, it may require a reconfigure of server-side hooks on docker hub if they exist for the current docker file though.

jbeezley:
The `Dockerfile` at the root actually needs to be there, but you can move the `/docker` directory anywhere.

jbeezley:
We don't currently have any automated docker deployments (dockerhub fails to build for some unknown reason).  Dockerfiles don't support `ADD`-ing files from parent directories, so we would need a manual step to copy the `Dockerfile` to the root.  I'm not even sure if that is possible in dockerhub.

zachmullen:
@aashish24 I moved the ansible and docker code under a "deploy" directory.

aashish24:
thanks @zachmullen. LGTM :+1:

zachmullen:
(BTW I don't have write access to this repo)

aashish24:
I will add you. In the mean time, I am merging this.

aashish24:
@doutriaux1 is it okay to add @zachmullen ?

doutriaux1:
of course.

PR #1492, opened 2015-08-13
get cmor via git repo

doutriaux1:
@dnadeau4 @aashish24 @sankhesh fetching cmor from github rather than tarball

### DIFF
--- a/CMake/cdat_modules/cmor_pkg.cmake
+++ b/CMake/cdat_modules/cmor_pkg.cmake
@@ -1,7 +1,9 @@
 set(CMOR_VERSION 2.9.2)
-set(CMOR_URL ${LLNL_URL})
-set(CMOR_GZ CMOR-${CMOR_VERSION}.tar.gz)
-set(CMOR_MD5 f7e78104878d93ffbd3cae2ccd95db2c)
+set(CMOR_BRANCH master)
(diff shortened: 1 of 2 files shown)


CMake/cdat_modules/cmor_pkg.cmake
```diff
@@ -1,7 +1,9 @@
 set(CMOR_VERSION 2.9.2)
-set(CMOR_URL ${LLNL_URL})
-set(CMOR_GZ CMOR-${CMOR_VERSION}.tar.gz)
-set(CMOR_MD5 f7e78104878d93ffbd3cae2ccd95db2c)
+set(CMOR_BRANCH master)
```
sankhesh: Instead of using the `master` branch, it would be better to set a specific commit as the SHA-1 sum. That way, changes in the upstream repository will not break UV-CDAT

doutriaux1: Doing the same as for vtk/vistrails/etc... Between releases we use master, then at release time we tag a specific sha1

aashish24:
@doutriaux1 @sankhesh sounds good to me. Having master for now is consistent.

aashish24:
:+1:

PR #1502, opened 2015-08-26
get the tests to pick up mac sample_data from correct place

doutriaux1:


testing/dv3d/TestManager.py
```diff
@@ -80,7 +80,7 @@ def __init__( self, name, **args ):
         self.image_name = os.path.join( self.test_dir, 'images', '.'.join( [ self.name, 'png' ] )  )
         filename = args.get( 'file', DefaultSampleFile )
         self.varnames = args.get( 'vars', [ DefaultSampleVar ] )
-        self.file_path = os.path.join( vcs.sample_data, filename )
+        self.file_path = os.path.join(*["/"]+vcs.__path__[0].split("/")[:-4]+["share","uvcdat","sample_data",filename] )
```
aashish24: @doutriaux1 why does the code in 83 not work on Mac? The new code seems too complicated.

doutriaux1: Yes, anyway the problem is deeper than that because nowit fails because some DV3D test need clt which is in another location... The right way to do this is to move geos5 file with other sample_data... Will do in the morning, lets kill that PR

doutriaux1: @aashish24 just to clarify setup.py puts the file in Library/Frameworks/Python/Version/2.7/share/uvcdat on mac. 
On Linux it is in share/uvcdat... There are two possibilities

* make our share/sample_data consistent with setup.py (different location between mac and linux)
* move @ThomasMaxwell sample data with our other sample data

I'm going to go with the later one.

PR #1507, opened 2015-08-29
Fix library problem when compiling python 2.7 and using VIM 7.4. Add …

dnadeau4:
…--enable-unicode=ucs4


doutriaux1:
oh thank you soooo much! Thats' been painfully hurting me for a while!

PR #1508, opened 2015-08-31
fix #959

### LINKED ISSUE #959: MV2.array(0.).fill_value is NaN!
In the `TransientVariable.__new__ method`, the following code assign a NaN value to the fill_value attribute:

```python

       fill_value = numpy.array(fill_value).astype(dtype),
```

So, for instance:

```python
>>> numpy.isnan(MV2.array(0.).fill_value)
True
```

doutriaux1:
new numpy set missing to nan by default we like 1.e20 better


doutriaux1:
@dnadeau4 @aashish24 please review

durack1:
@doutriaux1 this might be one of the fixes that gets us back up and running with `1.10.b1`..

Issue #1087, opened 2015-02-26
template.title.textorientation/table does not accept text object


doutriaux1:

doutriaux1:
works now in master added a test for it

PR #1509, opened 2015-08-31
Added test for #1087

doutriaux1:
@aashish24 just adding a test to make sure it never gets broken again. The problem reported by issue #1087 is actually working now.


aashish24:
Thanks @doutriaux1. One thing I don't understand is how you report an error if test fails? At the end it just seems to be listing the titles? Can you improve it a bit? May be use try and catch and return 1 or some other number to inform that it is failed indeed.

doutriaux1:
@aashish24 it use to not accept this types as input and we raise an error. So the test would fail.

aashish24:
Okay I trust your words @doutriaux1 

LGTM :+1:

PR #1510, opened 2015-08-31
fix #84

### LINKED ISSUE #84: add clear function to xmgrace module
Currently there is no function to clear the xmgrace window.

One can achieve clear a window by running:

```python
x('kill G0.S0')
```

on each graph/set

These combination can be obraimend via:

```python
sets = x.whichsets()
```

Probably would need to add functions to kill string and other objects as well.

Also there maybe an xmgrace command to do this already which would be the easiest way.

doutriaux1:


doutriaux1:
@aashish24 please review

Packages/xmgrace/Lib/xmgrace.py
```diff
@@ -3080,9 +3080,14 @@ def add_ellipse(self, x1, y1, x2, y2,
         self.nellipse = len(self.Box)
         return
 
+    def clear():
+        """ clears all graphs """
+        for i, g in enumerate(self.Graph):
+            for j in range(g.nset):
+                self("kill G%i.S%i" % (i, j))
```
aashish24: ??

/home/buildbot/buildbot/uvcdat-garant-linux-release/source/Packages/xmgrace/Lib/xmgrace.py:3085:31: F821 undefined name 'self'
        for i, g in enumerate(self.Graph):
                              ^
/home/buildbot/buildbot/uvcdat-garant-linux-release/source/Packages/xmgrace/Lib/xmgrace.py:3087:17: F821 undefined name 'self'
                self("kill G%i.S%i" % (i, j))
                ^
2       F821 undefined name 'self'

doutriaux1: yes I saw this last night it's already fixed, see latest commit

PR #1513, opened 2015-09-02
Add PYPTHONPATH explicitly to cmake to make sure the right CYTHON is …

dnadeau4:
…chosen.  There is a conflict on Ubuntu with Cython installed by the system.

Fix SciPy conflict with Cython installation from repository.

### DIFF
--- a/CMake/cdat_modules_extra/SCIPY_configure_step.cmake.in
+++ b/CMake/cdat_modules_extra/SCIPY_configure_step.cmake.in
@@ -1,6 +1,7 @@
 message("Configuring SCIPY:\n@SCIPY_binary@")
 set(ENV{@LIBRARY_PATH@} "@CMAKE_INSTALL_PREFIX@/lib:@cdat_EXTERNALS@/lib:$ENV{@LIBRARY_PATH@}")
 set(ENV{VS_UNICODE_OUTPUT} "")
+set(ENV{PYTHONPATH} "@PYTHONPATH@")
 
 # As explained in site.cfg.example - See http://projects.scipy.org/scipy/browser/trunk/site.cfg.example
 # Directories listed in variable such as library_dirs or source_dirs should be separated using os.pathsep
--- a/CMake/cdat_modules_extra/SCIPY_install_step.cmake.in
+++ b/CMake/cdat_modules_extra/SCIPY_install_step.cmake.in
@@ -1,6 +1,7 @@
 message("Installing SCIPY:\n@SCIPY_PREFIX_ARGS@")
 set(ENV{@LIBRARY_PATH@} "@CMAKE_INSTALL_PREFIX@/lib:@cdat_EXTERNALS@/lib:$ENV{@LIBRARY_PATH@}")
 set(ENV{VS_UNICODE_OUTPUT} "")
+set(ENV{PYTHONPATH} "@PYTHONPATH@")
 
 if(APPLE)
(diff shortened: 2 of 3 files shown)


aashish24:
thanks @dnadeau4. Should we set it here instead (In else): https://github.com/UV-CDAT/uvcdat/blob/master/CMake/cdat_modules/python_pkg.cmake

dnadeau4:
@aashish24 python_pkg.cmake is used to make python.  It is not needed at that level.  For a python package running setup.py we need to set PYTHONPATH to "%PYTHONPAHT%".  I learned that double quotes are important.

dnadeau4:
@doutriaux1 thanks!  I was just checking the "checks passed".

aashish24:
@dnadeau4 we include python.pkg regardless. So the idea I was proposing to set the environment there and then for everyother package PYTHONPATH would have been set vs us setting PYTHONPATH for individual packages. 

@doutriaux1 would have been nice to wait for the discussion to complete. I didn't approve the branch as we were still discussing.

aashish24:
https://github.com/UV-CDAT/uvcdat/blob/master/CMake/cdat_modules/python_pkg.cmake#L38

doutriaux1:
@aashish24 didn't see that you commented until after I merged :stuck_out_tongue_winking_eye: sorry about that. It's actually a good idea, probably worth adding in another branch. I'm worried it might break something else though, so it's better to do it in two separate branches.

aashish24:
No problem @doutriaux1 thanks for the consideration. It shouldn't break anything and if does then that would mean that we have some other issue / bug that we need to fix. I am just thinking that if SCIPY needs PYTHONPATH and it is likely that other python packages needs it too and may be that they are using system python and we are just lucky that we didn't run into any issue. 

+1 for pushing in separate branch

doutriaux1:
let's do it. I remember some user running into issues on mac where it would pick up the system numpy rather than ours this might hel pwith this (unreproducible on my end) issue.

dnadeau4:
Just came back from training.  If you think that adding PYTHONPATH to python_pkg.cmake would fix the problem for many packages, we should do it.  I can try it maybe tomorrow, today I have another meeting at 1:00pm to prepare.

dnadeau4:
@aashish24 I tried all afternoon to compile SCIPY modifying python_pkg but it fails every time.   I added these line on line 22

 * set(ENV{PYTHONPATH} "@PYTHONPATH@")
 * message("[INFO] ***** Python Path is @PYTHONPATH@")

cmake shows the following message
* [INFO] ***** Python Path is /home/dnadeau/uvcdat-latest/15-09-03/lib/python2.7/site-packages
  CMake Error at /software/build-uvcdat/CMake/SCIPY_configure_step.cmake:59
  (message):

 * Error in config of SCIPY
 
Same compilation error.
 * CMake Error at /software/build-uvcdat/CMake/SCIPY_configure_step.cmake:59 (message):
 * Error in config of SCIPY

Since we already have solution that work, I prefer to work on another more important issue.

aashish24:
@dnadeau4 thanks for trying this. No problem, I am surprised it didn't work. It could be that env is not getting passed to SCIPY for some reason. I will have a look at it but like you said this is not higher priority since we have  a solution in place. Thanks for looking into it.

dnadeau4:
It was worth a try,  I would have liked it to work.  Somehow the PYTHONPATH set in python_pkg is not passed when compiling SciPy.

PR #1520, opened 2015-09-06
Update g2clib to 1.4.0 and use libpng.

dnadeau4:
Need to update libcdms configure as well to add libpng to cdtime.so library path created by python.


aashish24:
@dnadeau4 looks like something is broken. Can you confirm if it is caused by your branch?

dnadeau4:
Yes it is!.   I am waiting to get push access to libcdms.   I have made some changes there that makes g2clib 1.4.0 works on MAC and Linux.

dnadeau4:
I will close the request until libcdms2 is pushed as well.

PR #1533, opened 2015-09-09
on the macbot it used to pick system png rather than XQuartz

doutriaux1:
so it couldn't import matplotlib._png
this should fix it


doutriaux1:
we have a green mac! No new failure on RH. Merging in!

Issue #1011, opened 2015-02-05
CMake warning that might break things eventually


doutriaux1:
```
CMake Warning (dev) at CMake/cmake_modules/pipinstaller.cmake:4 (include):
  Policy CMP0053 is not set: Simplify variable reference and escape sequence
  evaluation.  Run "cmake --help-policy CMP0053" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  For input:

    '@cdat_CMAKE_BINARY_DIR@/cdat_common_environment.cmake'

  the old evaluation rules produce:

    '/Users/doutriaux1/build/CMake/cdat_common_environment.cmake'

  but the new evaluation rules produce:

    '@cdat_CMAKE_BINARY_DIR@/cdat_common_environment.cmake'

  Using the old result for compatibility since the policy is not set.
Call Stack (most recent call first):
  CMake/cdat_modules/ipython_external.cmake:6 (include)
  CMakeLists.txt:631 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

PR #1536, opened 2015-09-10
Fixed cmake warnings and set appropriate policy

aashish24:
Ref: #1011

### DIFF
--- a/CMake/cmake_modules/ExternalProject.cmake
+++ b/CMake/cmake_modules/ExternalProject.cmake
@@ -156,6 +156,10 @@
 # (To distribute this file outside of CMake, substitute the full
 #  License text for the above reference.)
 
+# CMake policy usages
+cmake_policy(PUSH)
+cmake_policy(SET CMP0054 NEW)
+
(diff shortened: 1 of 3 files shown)


aashish24:
@doutriaux1  @sankhesh please review.

CMake/cmake_modules/ExternalProject.cmake
```diff
@@ -156,6 +156,10 @@
 # (To distribute this file outside of CMake, substitute the full
 #  License text for the above reference.)
 
+# CMake policy usages
+cmake_policy(PUSH)
+cmake_policy(SET CMP0054 NEW)
```
jbeezley: This policy was introduced in 3.1.  Probably should add `if(POLICY CMP0054)` to support older cmake's.

aashish24: Sure @jbeezley. This is a good point. I pushed a change.

doutriaux1:
error not related to this branch, messages are gone, merging in.

PR #1543, opened 2015-09-11
Dramatically cuts down on the number of renderers used per plot

chaosphere2112:
Hey all...

I was poking around in the latest off of master, and I bumped into a quite impressive regression in interactivity. Since I haven't touched `vtk_ui` in a while, I did some digging.

It looks like the regression was caused by the hatch/patterns PR; in particular, these are the troublesome lines:

https://github.com/UV-CDAT/uvcdat/blob/fc1660a30a6bdbafb5ee200cfeebd55f519ecb74/Packages/vcs/Lib/VTKPlots.py#L491-L529

which come from this commit:

https://github.com/UV-CDAT/uvcdat/commit/66465b5de8662c966a6049fc841c494d41ec11f0

The issue, as far as I can tell, is that this adds 300 `vtkRenderer`s to the render window (where previously there were 18). The main offender is the color bar, which apparently is rendered as a couple hundred fill areas with a pile of lines. All of these renderers dramatically increase the amount of time spent using `vtkPropPicker`, which is presumably the main cause of the slow down.

I made it so the primitives being generated will all use the same renderer for the many actors generated for that primitive. This brought the renderer count back down to like 27, which is a *lot* more reasonable than 300.

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -322,7 +322,7 @@ class VTKVCSBackend(object):
             self.renWin.Render()
 
     def createRenderer(self, *args, **kargs):
-            # For now always use the canvas background
+        # For now always use the canvas background
         ren = vtk.vtkRenderer()
         r, g, b = self.canvas.backgroundcolor
         ren.SetBackground(r / 255., g / 255., b / 255.)
@@ -492,6 +492,7 @@ class VTKVCSBackend(object):
             if gm.priority != 0:
                 actors = vcs2vtk.prepLine(self.renWin, gm)
                 returned["vtk_backend_line_actors"] = actors
+                create_renderer = True
                 for act, geo in actors:
                     ren = self.fitToViewport(
                         act,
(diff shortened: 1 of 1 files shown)


aashish24:
@chaosphere2112 thanks... but 27 is still a lot. Can you post a script that creates 27 renderer?

chaosphere2112:
It looks like I'm getting only 23 now, which is just 5 more than there were prior to the PR from @sankhesh. Here's a test script you can run with an older UV-CDAT as well as a current one to see the difference.

```
import vcs, cdms2
x = vcs.init()
try:
    f = cdms2.open(vcs.sample_data + "/clt.nc")
except AttributeError:
    # for older UV-CDATs
    f = cdms2.open(vcs.prefix + "/sample_data/clt.nc")
s = f("clt")
x.plot(s)
renderers = x.backend.renWin.GetRenderers()
renderers.InitTraversal()
ren_list = []
ren = renderers.GetNextItem()
while ren is not None:
    ren_list.append(ren)
    ren = renderers.GetNextItem()

print len(ren_list)
```

aashish24:
thanks @chaosphere2112 LGTM :+1:

doutriaux1:
@aashish24 did you check that the test passed?

aashish24:
Yup, always do. Some of the tests failing were not related to this branch (they were failing for some time on RH).

PR #1551, opened 2015-09-16
Skipping this test for RH6 as well

doutriaux1:


doutriaux1:
@dlonie @jbeezley @sankhesh @aashish24 any idea why these 3 test fail when enabling mesa?
https://open.cdash.org/viewTest.php?onlyfailed&buildid=4013750
I'm going to turn them off in case of RH/MESA

aashish24:
@doutriaux1 it seems to that on this machine volume rendering is not working. What version of MESA are you using on this system?

durack1:
@aashish24:
```
[user@ocean ~]$ glxinfo | grep "OpenGL version"
OpenGL version string: 2.1 Mesa 10.4.3
```

aashish24:
Recently I found out that Mesa 10.5.5 or later is better. Is this Mesa installed by our superbuild?

jbeezley:
I have seen the same issue with mesa on Ubuntu as well.

doutriaux1:
We are installing 7.1 should I try to upgrade?

jbeezley:
Someone (maybe @cjh1?) told me a while back that only specific versions of mesa work with VTK.  I have never tried though.

durack1:
@aashish24 nope, looks like 7.6.1:

https://github.com/UV-CDAT/uvcdat/blob/master/CMake/cdat_modules/osmesa_pkg.cmake#l6-8

aashish24:
It used to be 7.* some version and while it may be still true for VTK OpenGL 1 backend, 10.5.5. is minimum for OpenGL 2 version (which we should switch to in 6 months - year). If we can try 10.5.5 for OpenGL 1 version of VTK (which we have right now), it would be great.

doutriaux1:
got it to pass all test except the one from BUILD_MODE=ALL, merging.

PR #1553, opened 2015-09-18
adds function on colormap objects to set RGB values

### LINKED ISSUE #1484: colormap need user level func to set colorindices
Bill Hayes reported:
```python
v = vcs.init()
cm = v.createcolormap('cm', 'default')
cm.color = 0, 99,99,99     # does NOT set the color  (but does not error)
```
However the following works to set and get values:
```python
cm.index[0] = [99,99,99]   # this works
cm.index[0]    # returns the RGB list
```
THe user here is using internal/private functions
The correct call is
```python
v.setcolormap(cm.name)
v.setcolorcell(0,99,99,99)
```

However, since 2.2 graphic methods are allowed to have their own colormap so we should have actual function on the colormap object to set the colors (rather that the private ones only)

doutriaux1:
fix #1484


testing/vcs/test_vcs_set_cmap_colorcell.py
```diff
@@ -0,0 +1,10 @@
+import vcs
```
aashish24: Instead of making two tests.. can you just make one? We have lot many tests and now we need to make sure that we don't grow the number as much

doutriaux1: I like one test per issue. So we know exactly what broke.

aashish24: @doutriaux1 like I said, we are having two many tests. In this case, I think you can easily put both of them together into one as they are testing getter and setters.

doutriaux1: no, we have established a policy of one test per feature. It takes 2 seconds at most to run, this will save one second.

aashish24:
per feature not per getter and setter. Also, @williams13 bought this issue as well. Even if you are writing unit tests, you can include them in one file vs having each into own separate file. The main thing is you are testing the code (the lines). Having a very large number of test files is not good either.

aashish24:
I am just suggesting best software practices. I understanding if you are testing a feature (math, model of some sort etc) then yes, having individual tests help.

doutriaux1:
@jbeezley paranthesis closed in add_test too early. Pushing a fix now.

jbeezley:
In this particular case, I think the tests could be combined into one, but in general I don't think there is such a thing as "too many tests" in any practical sense.  There is a point at which testing can become too slow, but I would argue it being a result of either a bad unit testing framework (i.e. lots of setup code, not mocking process intensive components) or a fundamental problem with software itself (i.e. orthogonal components cannot be tested independently).

doutriaux1:
@aashish24 go ahead if you have time, push an update to the branch.

aashish24:
> In this particular case, I think the tests could be combined into one, but in general I don't think there is such a thing as "too many tests" in any practical sense. There is a point at which testing can become too slow, but I would argue it being a result of either a bad unit testing framework (i.e. lots of setup code, not mocking process intensive components) or a fundamental problem with software itself (i.e. orthogonal components cannot be tested independently).

@jbeezley I agree with most of you said. Like I said earlier, unnecessarily increasing number of test files is not a good practice unless required. If we can combine tests into one file that would be great. This is particularly true for unit tests which is pretty much what we have in this branch. More testing is good though. I didn't mean that. We should continue to add more tests to increase code coverage.

aashish24:
Will do later today @doutriaux1 thanks.

jbeezley:
I don't think the number of test files is a useful metric for anything.  My point is that, assuming identical coverage, more tests is better than fewer tests.

aashish24:
Agree @jbeezley :+1: We are on the same page.

doutriaux1:
@aashish24 so you are on the same page as @jbeezley two files are better than one :wink: 

>I don't think the number of test files is a useful metric for anything. My point is that, assuming identical coverage, more tests is better than fewer tests.

aashish24:
> In this particular case, I think the tests could be combined into one

Jon said that in this particular case one is better (I was saying the same). One is better in this case, in some other cases two or more is desired.

doutriaux1:
@aashish24 any idea what happened on crunchy?

aashish24:
@doutriaux1 no idea. Its only one test failing on crunchy... I think the changes are safe enough so I am willing to merge it.

aashish24:
LGTM :+1:

PR #1556, opened 2015-09-18
Issue 1396 bad png path

doutriaux1:


Packages/vcs/Lib/Canvas.py
```diff
@@ -4812,6 +4812,9 @@ def png(self, file, width=None, height=None,
     a.plot(array)
     a.png('example')       # Overwrite a png file
 """
+        base = os.path.dirname(file)
+        if base!="" and not os.path.exists(base):
+          raise vcsError("Output path: %s does not exists" % base)
```
durack1: @doutriaux1 "does not exists" -> "does not exist"

Packages/vcs/Lib/Canvas.py
```diff
@@ -4812,6 +4812,9 @@ def png(self, file, width=None, height=None,
     a.plot(array)
     a.png('example')       # Overwrite a png file
 """
+        base = os.path.dirname(file)
+        if base != "" and not os.path.exists(base):
+            raise vcsError("Output path: %s does not exists" % base)
```
durack1: @doutriaux1 "does not exists" -> "does not exist"

Packages/vcs/Lib/Canvas.py
```diff
@@ -4812,6 +4812,9 @@ def png(self, file, width=None, height=None,
     a.plot(array)
     a.png('example')       # Overwrite a png file
 """
+        base = os.path.dirname(file)
```
aashish24: Why not put this block of code in try catch instead of checking of the file exists? In general thats a better practice

doutriaux1: Because it does not fail, that is actually the issue.

doutriaux1: @aashish24 are you suggesting we create the subdirectory?

Packages/vcs/Lib/Canvas.py
```diff
@@ -4812,6 +4812,9 @@ def png(self, file, width=None, height=None,
...
+        if base != "" and not os.path.exists(base):
+            raise vcsError("Output path: %s does not exist" % base)
         return self.backend.png(
```
aashish24: Do you mean that backend.png does not throw an exception or any other error/

Packages/vcs/Lib/Canvas.py
aashish24: No, I was not.

aashish24:
LGTM :+1:

PR #1557, opened 2015-09-18
Fix view rotation when rendering in background mode

sankhesh:
Fixes #1446

PR #1540, opened 2015-09-10
issue1538 durack1 update Cython 0.22.1 to 0.23.3

durack1:
Fix #1538 - Update Cython 0.22.1 to 0.23.1

Issue #653, opened 2014-09-16
The vectors are not drawn when using projection other than plat-caree.


aashish24:

dlonie:
I'm looking into this, and it's difficult to figure out what data is in what coordinate system, what coordinate system it should be in, etc.

I'm using the test command:

```python "/ssd/src/uvcdat/testing/vcs/test_vcs_basic_gms.py" "--gm_type=vector" "--projection=-3" "--lat1=-90" "--lat2=0" "--source=/ssd/src/uvcdat-build/uvcdat-testdata/baselines/vcs/test_vcs_basic_vector_-3_proj_SH.png" --show```

Prior to calling doWrap on the vector data, the vector data bounds are similar to the final continent bounds (Roughly `(-9782112.62778627, 9772957.562484365, -9759498.0, 8568893.0)`). After calling doWrap, the vector dataset is empty, as there is no data that lies within the clip box.

If there is a missing projection, where should that projection take place? Should each worker function (plotVector, plot2D, plot[...], etc) take care of projecting the data, or should this be handled somewhere higher up the stack?

I guess what I'm looking for is sort of a high level description of how data comes into vcs (e.g. what coordinate system), and what transformations are expected to take place, what order the transformations should be performed in, and where the transformations are expected to occur.

doutriaux1:
@dlonie I think you're onto something. doWrap maniuplates the inputData ut in the case of vectors we are not using this we are using some other array. I can we need to update  doWrap to also alter the "Vectors array"

dlonie:
I'll reassign this as it's a VCS issue, and you have a better understanding of what that code should be doing ;)

doutriaux1:
@dlonie taking out doWrap does make any diff the bounds are big as soon as I do the geo (projection). The numbers are probably ok too, they are in m on the earth sphere. I just think the fitToViewPort might be the issue here. Not sure.

doutriaux1:
when fixing this we should turn back on vector tests for projections

doutriaux1:
@aashish24 @dlonie is one f you taking a look at this?

dlonie:
@doutriaux1 I haven't looked at it since it was reassigned.

doutriaux1:
``python
import vcs
import os
import cdsm2
f=cdms2.open(os.path.join(vcs.sample_data,"clt.nc"))
u=f("u")
v=f("v")
V=x.createvector()
p=x.createprojection()
p.type="polar"
V.projection=p
[...]

PR #1560, opened 2015-09-19
Fix vector plots

aashish24:
#### WIP: I have observed tests failing on my machine but the difference seems small. I just wanted to see how they run on buildbot machine.


aashish24:
Ref: #653

doutriaux1:
@aashish24 still lots of failures here. What changed?

aashish24:
@doutriaux1 thats puzzling me. On my machine, I am getting only vector tests failure. Although I am using the patched VTK

aashish24:
@doutriaux1 still the same. For some reason, I am not seeing these failing tests on my machine.

Packages/vcs/Lib/vcsvtk/vectorpipeline.py
```diff
@@ -27,14 +30,45 @@ def plot(self, data1, data2, tmpl, grid, transform):
         data1 = self._context().trimData2D(data1)
         data2 = self._context().trimData2D(data2)
 
+        scale = 1.0
+
+        lat = data1.getLatitude()[:]
```
sankhesh: data1 could be `NoneType` 

`vcs_no_xtra_elts` fails because of that. (https://open.cdash.org/testDetails.php?test=375905204&build=4024522)

aashish24:
@doutriaux1 this branch should be good to now. The problem (failing tests) was with VTK bug. I will fix the doWrap method in another PR.

aashish24:
Goes with https://github.com/UV-CDAT/uvcdat-testdata/pull/67/files

aashish24:
```
[] Fix doWrap 
```

doutriaux1:
@aashish24 waiting for the bots to be done to confirm merge. Thanks for pushing this in.

doutriaux1:
@aashish24 4 failures on oceanonly, but I think it is because master was merged in by bots.

aashish24:
Right. there was an issue with the test-data. The only real test failing is test_no_extra_elements

aashish24:
@doutriaux1 most of the tests are passing now. The multiple format one and label background one is failing for some other reason.

aashish24:
@doutriaux1 this branch is looking good now. Not sure why the label and the click test is failing. 

https://open.cdash.org/testDetails.php?test=376700167&build=4028520

durack1:
@aashish24 the reason is due to problems with the test image, so compare the baseline:
test_vcs_click_info
With the test:
test
You'll see that there is bleeding of color outside the box, and the y-axis labels in particular are offset, look at 90N.. There are also problems with the x-axis labels and their placement too..

aashish24:
@doutriaux1 could it be caused by my changes? thanks @durack1

durack1:
@aashish24 I've been seeing this issue in a number of PRs #1540, #1560, #1557 - and it only seems to be a problem on `garant` - so an `OS-specific` issue?

durack1:
@aashish24 @doutriaux1 I'm wondering if `garant` is just not syncing correctly with the updated source and as it's behind we're getting these failures - looks like the case with #1564..

jbeezley:
It doesn't merge in master like the others.  Is that the issue?

durack1:
@jbeezley it seems that in particular with #1564 `FFmpeg` issues are only occurring on `garant`, whereas everything else passes.. So like it fails to pull across all the PR changes or something? Dunno?

aashish24:
thanks @durack1 for the information. In that case, I think this branch looks good to me. Can someone review it @durack1 @sankhesh @jbeezley @chaosphere2112

testing/vcs/test_vcs_vectors_robinson.py
```diff
@@ -0,0 +1,24 @@
+import vcs, cdms2, numpy, os, sys
```
sankhesh: This file doesn't comply with PEP8 style.

aashish24: @sankhesh but I don't see flake8 test failing. what I am missing?

jbeezley: We probably don't check the test scripts.

aashish24: Ah I see. Thanks @jbeezley. I will push a fix to follow the PEP8 stype in my next branch as I need to take care of doWrap as well.

aashish24:
thanks @doutriaux1

PR #1561, opened 2015-09-21
Add windfield dependency to rpy2 getting rid of parallel installation…

dnadeau4:
… issue.


doutriaux1:
yoohoo! RH FULL passes!

Issue #1586, opened 2015-10-06
New libx264/h264 library bug needs fix


durack1:
@chaosphere2112 I've just tried to roadtest the new `libx264/h264` defaults in the `ffmpeg` call and hit the following error:

So the call:
```python
x.ffmpeg(outMP4,xList,rate=5,bitrate=2048)
```

returns:
```python
** Processing:  /work/durack1/Shared/131118_PaperPlots_SaltyVariability/pngs/quarters.mp4                                            
ffmpeg version 2.7 Copyright (c) 2000-2015 the FFmpeg developers                                                                     
  built with gcc 4.4.7 (GCC) 20120313 (Red Hat 4.4.7-16)                                                                             
  configuration: --prefix=/usr/local/uvcdat/2.4.0rc1/Externals --disable-yasm --enable-gpl --enable-libx264 --extra-cxxflags=/export/doutriaux1/build/build/FFMPEG --enable-shared --enable-zlib                                                                          
  libavutil      54. 27.100 / 54. 27.100                                                                                             
  libavcodec     56. 41.100 / 56. 41.100                                                                                             
  libavformat    56. 36.100 / 56. 36.100                                                                                             
  libavdevice    56.  4.100 / 56.  4.100                                                                                             
  libavfilter     5. 16.101 /  5. 16.101                                                                                             
  libswscale      3.  1.101 /  3.  1.101                                                                                             
  libswresample   1.  2.100 /  1.  2.100                                                                                             
  libpostproc    53.  3.100 / 53.  3.100                                                                                             
Input #0, image2, from '/export/durack1/.uvcdat/__uvcdat_309035998_%d.png':                                                          
  Duration: 00:00:20.00, start: 0.000000, bitrate: N/A                                                                               
    Stream #0:0: Video: png, rgb24(pc), 814x629, 5 fps, 5 tbr, 5 tbn, 5 tbc                                                          
[libx264 @ 0x1a1c9a0] height not divisible by 2 (814x629)                                                                            
Output #0, mp4, to '/work/durack1/Shared/131118_PaperPlots_SaltyVariability/pngs/quarters.mp4':                                      
    Stream #0:0: Video: h264, none, q=2-31, 128 kb/s, 5 fps                                                                          
    Metadata:                                                                                                                        
      encoder         : Lavc56.41.100 libx264                                                                                        
Stream mapping:                                                                                                                      
  Stream #0:0 -> #0:0 (png (native) -> h264 (libx264))                                                                               
[...]

PR #1592, opened 2015-10-06
x264 fixes

chaosphere2112:
Fixes a few bugs...

#1586 
@doutriaux1's issue with expanding x264 on a newer CMake
@dnadeau4's issue with building x264 on his VM


CMake/cdat_modules/x264_pkg.cmake
```diff
@@ -1,6 +1,6 @@
-set(X264_DATE 20150921)
+set(X264_DATE 20151006)
 set(X264_TIME 2245)
-set(X264_ADDENDUM "-stable")
+set(X264_ADDENDUM "")
 set(X264_URL ${LLNL_URL})
 set(X264_GZ x264-snapshot-${X264_DATE}-${X264_TIME}${X264_ADDENDUM}.tar.gz)
 set(X264_MD5 43095e5d2ca331eec7f0fbed856e5bab)
```
doutriaux1: no new md5?

doutriaux1:
@chaosphere2112 yep that's what I was suspecting you probably need to update the md5

chaosphere2112:
@doutriaux1 @dnadeau4 Sorry about that, just wanted to push the PR out before I went home for the day. Fixed both of those issues.

CMake/cdat_modules/x264_deps.cmake
```diff
@@ -1 +1 @@
-set(X264_DEPS "")
+set(X264_deps ${pkgconfig_pkg})
```
aashish24: So X264_deps just depends on pkgconfig? Are there any other system specific libraries we need to document?

chaosphere2112: Nope; in fact, it technically doesn't even depend on `pkgconfig`. It has a soft dependency on `yasm`, but that's already sort of a dependency for us (as part of the original x264 PR I did some stuff to make our build handle not having `yasm` installed).

chaosphere2112: The reason it's "dependent" on pkgconfig is because it fixes a weird parallelism issue that @dnadeau4 was having (where the linker got confused as `libx264` and `pkgconfig` built at the same time).

aashish24: thanks @chaosphere2112.care to add a note in the deps file so that someone know it next time?

aashish24:
No problem @chaosphere2112 thanks for addressing our suggestions. Once cdash is happy, I will give it +2

chaosphere2112:
@doutriaux1 Those have been stuck in queue an awfully long time... any idea on the status?

doutriaux1:
I killed the server, I need to restart it.

PR #1602, opened 2015-10-09
Fixed doWrap method for vectors

aashish24:


aashish24:
Fixes #653 #1560

aashish24:
@doutriaux1 @sankhesh please review. 

@doutriaux1 how can I create a test that will force the code to duplicate the data?

aashish24:
@williams13 if this works, this should fix all of the remaining issues with vector plots and projection. Please test it when you get a chance

williams13:
Will do!  Thanks!

aashish24:
@doutriaux1 all of the tests are passing.

doutriaux1:
@aashish24 please add a test for it. How do you even know it works if you didn't test it?
to test read data let's say between 0 360 and plot -180/180 or vice versa.
to read data within a range:
```python
import cdms2
f=cdms2.open(file)
s=f(data,longitude=(0,360))
```

doutriaux1:
@aashish24 also your commit quality wasn't what you used us to. No description of what you changed and why. A bit disappointed here :wink:

doutriaux1:
@aashish24 I just add a test that shows it is still broken. Run the file with either `lon1 = 0.` or `lon1 = -180.` it fails in two different ways.

aashish24:
> @aashish24 also your commit quality wasn't what you used us to. No description of what you changed and why. A bit disappointed here

:) The changes I made should be obvious and since the other lines were not commented, I didn't comment new lines. I am planning to update these classes in the future.

aashish24:
@doutriaux1 ah.. thanks for the test. that was my other question.

aashish24:
@doutriaux1 your original test code is buggy, you are getting clt for both u and v. If I modified your code to get the actual vectors, I am getting the right plot. Here is the code:
```python
import cdms2
import vcs
import os
import sys

f=cdms2.open(os.path.join(vcs.sample_data,"clt.nc"))

# -180 failks as well
lon1 = -180
u=f("u")
v=f("v")

u=u(longitude=(lon1,lon1+360.))
v=v(longitude=(lon1,lon1+360.))

x=vcs.init()
Vec = x.createvector()
p = x.createprojection()
p.type = "mercator"
Vec.projection = p
x.plot(u,v,Vec)
raw_input("press enter")
```
mercator

aashish24:
Now I used the mercator. With Polar, I think vector plot will be really bad in general but I think that's a separate issue which I believe I mentioned earlier that vector plot in polar coordinates may require some more work.

aashish24:
@doutriaux1 fixed the scaling issue that was causing vectors not to show up in polar coordinates.

aashish24:
Goes with: https://github.com/UV-CDAT/uvcdat-testdata/pulls

aashish24:
@doutriaux1 @sankhesh please approve

doutriaux1:
@aashish24 my original code was NOT buggy, "u" and "v" are wrong in this file and do not get wrapped around by cdms2. Which is why I used "clt" twice, we don't care about how realstic the test is, only that is reproduces the features we are trying to fix.

doutriaux1:
@aashish24 you need to at least flake8 it. I will review your test because I'm worried your change to "u" and "v" actually prevents it from reproducing the bug we are trying to fix.

aashish24:
@doutriaux1 just using CLT twice would not produce any reasonable vectors. I don't understand why we cannot use u and v? for all other tests aren't we using u and v? I was able to reproduce the bug if I don't fix the scale factor. In case of POLAR coordinates the scales were off.

aashish24:
@doutriaux1 I will fix the style issue. I didn't see them on garant and that's why I thought all is good.

aashish24:
> @aashish24 my original code was NOT buggy, "u" and "v" are wrong in this file and do not get wrapped around by cdms2. Which is why I used "clt" twice, we don't care about how realstic the test is, only that is reproduces the features we are trying to fix.

@doutriaux1 thanks. Is it possible to use better data? If we use clt twice then when projected, vectors will be drawn really bad or not drawn at all.

aashish24:
@doutriaux1 flake8 should be fixed now (had unused import). Also, I tried with clt (for both u and v) works for me. Once you look at it, I can update the test to use clt

aashish24:
Okay, tested with clt, vectors look awful but the code seems to be working. @doutriaux1 please let me know if the code is still broken. I saw another issue (not with vectors but with continental lines).

aashish24:
@doutriaux1 should be all good. Please review.

aashish24:
@doutriaux1 ping!

aashish24:
@doutriaux1

aashish24:
@doutriaux1 :smiley_cat:

doutriaux1:
@aashish24 catching up... slowly...

doutriaux1:
@aashish24 still fails for ploar is that expected you say?

doutriaux1:
polar

aashish24:
@doutriaux1 if you use just clt and polar, yes since it will spread them weirdly. If i just use u and v (no wrapping) it does look reasonable but please have a closer look at polar one. All other projections looks great to me.

aashish24:
@doutriaux1 thanks :+1: :smile: :

PR #1605, opened 2015-10-12
Fix pattern 9

chaosphere2112:
This PR substantially speeds up pattern 9 (no longer takes most of a minute!). It also refactors the patterns/hatches to be classes, which will let us do some cool stuff later down the line (like custom patterns) and makes it so the logic for fetching a specific pattern is less hardcoded (again, making it easier to add custom patterns later). Matching testdata PR is at UV-CDAT/uvcdat-testdata#73


chaosphere2112:
@sankhesh You'll want to take a peek to make sure I didn't mess anything up in here.

doutriaux1:
@chaosphere2112 darn your baseline somehow dosen't let master baseline be merged into it. I wonder why. Did you start the baseline branch a while back?

aashish24:
:clap: : @chaosphere2112 what was the reason for the slow down?

chaosphere2112:
@aashish24 `FillTube` is kind of slow (.003 seconds per call), and Pattern 9 called it 4 times per 16 pixel square. I made it so instead of drawing it like this:

```
_______
|   \    /    |
|   /    \    |
------------
```

it is instead drawn like so:

```
_______
|   | _ |   |
|   | _ |   |
------------
```

and then rotated 45 degrees. I was able to dramatically reduce the number of draw calls (1 per column and row, and 1 per intersection) as well as using the speedier `FillBox` function.

aashish24:
thanks @chaosphere2112 so how did you trace FillTube timings? Also, thanks for changing the name from draw to paint. Actually "create" or "construct" would have better, but I am okay with paint.

chaosphere2112:
I just used `cProfile` and sorted by `tottime`

aashish24:
> I just used cProfile and sorted by tottime

Thanks, it would be nice to document developer tips somewhere...

aashish24:
@sankhesh the changes looks good to me but you are the original author. Please review.

chaosphere2112:
@aashish24 Documentation for that is pretty much the top google result for "profile python" :wink:

aashish24:
> @aashish24 Documentation for that is pretty much the top google result for "profile python"

:laughing: sure, I meant in a large sense.. not just for this. We need more extensive developers documentation.

sankhesh:
:+1:

PR #1617, opened 2015-10-16
fix #1281 default marker size too small

### LINKED ISSUE #1281: Cosmetic Bug: dots in Scatter plot are hardly visible.
Plot Scatter with u and v. Dots are hardly visible on smaller screen.

screen shot 2015-05-05 at 9 01 47 pm

Maybe we should change the color of the dots or the size.
I am not sure of labels, please remove if its unrelated.

doutriaux1:
also let the marker size by any positive number so if users find it too big they can make it smaller

### DIFF
--- a/Packages/vcs/Lib/vcs2vtk.py
+++ b/Packages/vcs/Lib/vcs2vtk.py
@@ -1214,6 +1214,7 @@ def prepGlyph(g, marker, index=0):
     if t == 'dot':
         gs.SetGlyphTypeToCircle()
         gs.FilledOn()
+        s *= numpy.pi
     elif t == 'circle':
         gs.SetGlyphTypeToCircle()
         gs.FilledOff()
--- a/Packages/vcs/Lib/marker.py
+++ b/Packages/vcs/Lib/marker.py
@@ -229,14 +229,14 @@ class Tm(object):
         return self._size
 
     def _setsize(self, value):
-        if isinstance(value, int):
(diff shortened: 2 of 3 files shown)


doutriaux1:
@aashish24 @sankhesh @chaosphere2112 after the bots are done please review
goes with: https://github.com/UV-CDAT/uvcdat-testdata/pull/76

Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -1214,6 +1214,7 @@ def prepGlyph(g, marker, index=0):
     if t == 'dot':
         gs.SetGlyphTypeToCircle()
         gs.FilledOn()
+        s *= numpy.pi
```
aashish24: Why not make it a integer number say 10, 20? it may help if we are going to do pixel align rendering.

doutriaux1: I was looking at numbers with a scientist, 3 wasn't right and 4 too big. So I made it fun. Beside s is not an integer to start with.

aashish24: > I was looking at numbers with a scientist, 3 wasn't right and 4 too big. So I made it fun. Beside s is not an integer to start with.

I see. I am bit surprised that s is not an integer.

aashish24:
other than that it LGTM :+1:

PR #1618, opened 2015-10-16
Made ESMF fail the build when fortran is the wrong version

chaosphere2112:
Does what it says on the tin.


CMake/cdat_modules/esmf_pkg.cmake
```diff
@@ -41,9 +41,7 @@ if(CDAT_BUILD_ESMF_ESMP)
         ## On APPLE need to test for -arch as well!
         add_cdat_package(ESMF "" "Build ESMF" ON)
       else()
-        message("[INFO] Skipping ESMF")
-        message("[INFO] gfortran version needs to be at least 4.3 to install ESMF")
-        message("[INFO] You have ${Fortran_VERSION}")
+        message(FATAL_ERROR "[ERROR] gfortran must be 4.3 < version < 5.0; you have ${Fortran_VERSION}")
```
aashish24: I think it should be 

gfortran must be 4.3 <= version < 5.0

aashish24:
LGTM :+1:

aashish24:
thanks @chaosphere2112

Issue #1620, opened 2015-10-19
XvsY error


chaosphere2112:
```
import cdms2, vcs
x = vcs.init()
f = cdms2.open(vcs.sample_data + "/clt.nc")
x.xvsy(f("v"), f("clt"))
```

gives:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-b691057b570d> in <module>()
----> 1 x.xvsy(v, s)

/Users/fries2/build-uvcdat/install/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.pyc in xvsy(self, *args, **parms)
   1705 """
   1706         arglist = _determine_arg_list('xvsy', args)
-> 1707         return self.__plot(arglist, parms)
   1708     xvsy.__doc__ = xvsy.__doc__ % (plot_keywords_doc,
   1709                                    graphics_method_core,

/Users/fries2/build-uvcdat/install/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/vcs/Canvas.pyc in __plot(self, arglist, keyargs)
   3498                 elif tp == "default":
   3499                     tp = "boxfill"
-> 3500                 gm = vcs.elements[tp][arglist[4]]
   3501                 if hasattr(gm, "priority") and gm.priority == 0:
   3502                     return

KeyError: '__xvsy_653912073685959_xvsy_'
```

doutriaux1:
Actually this shouldn't work because the shape won't match, but it's definitely crashing first for another reason.

williams13:
No it shouldn’t work, but it shouldn’t crash either. Just doing typical user things…

doutriaux1:
@chaosphere2112 your example at the top still fails for me in master.

doutriaux1:
```
  File "/git/uvcdat/testing/vcs/test_vcs_1d_fails_2_slabs.py", line 5, in <module>
    x.xvsy([1,2,3,4],[1,2,3,4],bg=False)
  File "/lgm/uvcdat/2015-11-09/lib/python2.7/site-packages/vcs/Canvas.py", line 1707, in xvsy
    return self.__plot(arglist, parms)
  File "/lgm/uvcdat/2015-11-09/lib/python2.7/site-packages/vcs/Canvas.py", line 3500, in __plot
    gm = vcs.elements[tp][arglist[4]]
KeyError: '__1d_64410090437423'
```

[...]

PR #1621, opened 2015-10-19
Fixed issue where XvsY was using the wrong element name

chaosphere2112:
Resolves issue #1620

### DIFF
--- a/Packages/vcs/Lib/manageElements.py
+++ b/Packages/vcs/Lib/manageElements.py
@@ -843,7 +843,7 @@ vcs.show('xvsy')
         source = source[:-6]
     name, source = check_name_source(name, source, 'xvsy')
     gm = unified1D.G1d(name + "_xvsy_", source + "_xvsy_")
-    vcs.elements["xvsy"][name] = gm
+    vcs.elements["xvsy"][name + "_xvsy_"] = gm
     return gm
 createxvsy.__doc__ = createxvsy.__doc__ % (
     plot_keywords_doc, graphics_method_core, axesconvert, create_GM_input, xvsy_output)
--- a/Packages/vcs/Lib/manageElements.py
+++ b/Packages/vcs/Lib/manageElements.py
@@ -25,11 +25,12 @@ from xmldocs import plot_keywords_doc, graphics_method_core, axesconvert,\
     scatter_output
 import random
 from error import vcsError
+import warnings
 import dv3d
 
 
 def check_name_source(name, source, typ):
-    """makes ure it is a unique name for this type or generates a name for user"""
+    """make sure it is a unique name for this type or generates a name for user"""
     elts = vcs.listelements(typ)
     if name is None:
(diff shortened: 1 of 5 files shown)


chaosphere2112:
Apparently this applies to all 1D plots that are done via the canvas methods. I'm going to fix them all.

williams13:
Thanks Sam!

chaosphere2112:
@doutriaux1 @williams13 

```
xvsy
yxvsx
xyvsy
scatter
```

Looks like that's all of the ones in `manageElements`; should be good now?

doutriaux1:
Ok should we make it exit nicely in that PR or should we create a separate issue for this?

chaosphere2112:
I think that should be in a separate issue; this one is pretty cut and dry.

doutriaux1:
done at: https://github.com/UV-CDAT/uvcdat/issues/1622

doutriaux1:
@chaosphere2112 it seems to have broken: vcs_verify_remove_objects

chaosphere2112:
OK, so I tweaked how we store 1D objects in `vcs.elements`. They're now exclusively stored in `vcs.elements["1d"]`. On a name collision, it will attempt to create a new one called `name_xyvsy` and issue a warning that the GM will have a different name than the one provided. When retrieving `name`, it will call `get1d(name)`. If that returns a graphics method, it checks the `g_type` property on the object, which returns the 1D type based on the attributes of the object (`xyvsy` has `flip` set to `True`, `scatter` has `linewidth` set to `0`). If the `g_type` doesn't line up with what we're expecting, it'll try and retrieve `name_xyvsy` (so users who don't see the warning should hopefully have everything happen correctly).

`isyxvsx` and `isxvsy` will now (possibly confusingly) claim that one graphics method could be either. This is true from the graphics method standpoint; `xvsy` and `yxvsx` are only different with regards to the number of slabs passed, which means that from this layer of the codebase, they're functionally identical. This is why I updated the `test_queries` test.

@aashish24 @doutriaux1

doutriaux1:
Awesome! Thanks @chaosphere2112

chaosphere2112:
@doutriaux1 Any idea what's up with the failed tests on crunchy?

doutriaux1:
it says
```
ERROR: In /export/doutriaux1/uvcbot/build/build/VTK/Rendering/OpenGL/vtkXOpenGLRenderWindow.cxx, line 1473
vtkXOpenGLRenderWindow (0x2e19f80): bad X server connection. DISPLAY=:5. Aborting.
```
probably need to restart `Xvfb`

chaosphere2112:
@doutriaux1 @aashish24 So are we good to merge, then?

aashish24:
@chaosphere2112 looks good to me if @doutriaux1 is okay with this change. 

:+1:

chaosphere2112:
@doutriaux1 @aashish24 ping

PR #1647, opened 2015-10-29
Fix compilation on Ubuntu with gcc 5.2.1

dnadeau4:
testing if "ncurses" patch works on different machines.

### DIFF
--- a/CMake/cdat_modules/esmf_pkg.cmake
+++ b/CMake/cdat_modules/esmf_pkg.cmake
@@ -37,11 +37,11 @@ endif()
 if(CDAT_BUILD_ESMF_ESMP)
     if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL GNU)
       # GNU gfortran must be >= 4.3
-      if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 2)
+      if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 1)
         ## On APPLE need to test for -arch as well!
         add_cdat_package(ESMF "" "Build ESMF" ON)
       else()
-        message(FATAL_ERROR "[ERROR] gfortran must be 4.3 <= version < 5.0; you have ${Fortran_VERSION}")
+        message(FATAL_ERROR "[ERROR] gfortran must be 4.1 <= version < 5.2; you have ${Fortran_VERSION}")
       endif()
     else()
       add_cdat_package(ESMF "" "Build ESMF" ON)
--- a/CMake/cdat_modules/esmf_pkg.cmake
+++ b/CMake/cdat_modules/esmf_pkg.cmake
@@ -36,12 +36,12 @@ endif()
 
 if(CDAT_BUILD_ESMF_ESMP)
     if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL GNU)
-      # GNU gfortran must be >= 4.3
-      if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 1)
+      # GNU gfortran must be >= 4.3 last 4.2 gfortran release was 4.2.4
+      if(${Fortran_VERSION} VERSION_GREATER "4.2.9" AND ${Fortran_VERSION} VERSION_LESS "5.2.2")
         ## On APPLE need to test for -arch as well!
         add_cdat_package(ESMF "" "Build ESMF" ON)
       else()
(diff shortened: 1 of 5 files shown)


CMake/cdat_modules/esmf_pkg.cmake
```diff
@@ -37,11 +37,11 @@ endif()
 if(CDAT_BUILD_ESMF_ESMP)
     if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL GNU)
       # GNU gfortran must be >= 4.3
-      if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 2)
+      if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 1)
```
doutriaux1: That is odd, I thought @chaosphere2112 had made it so that we have to have less than 5, plus 4.0 would fail with this no?

CMake/cdat_modules/esmf_pkg.cmake
```diff
@@ -37,11 +37,11 @@ endif()
...
         ## On APPLE need to test for -arch as well!
         add_cdat_package(ESMF "" "Build ESMF" ON)
       else()
-        message(FATAL_ERROR "[ERROR] gfortran must be 4.3 <= version < 5.0; you have ${Fortran_VERSION}")
+        message(FATAL_ERROR "[ERROR] gfortran must be 4.1 <= version < 5.2; you have ${Fortran_VERSION}")
```
doutriaux1: I thought fortran 5.x was not possible?

doutriaux1:
I have a question about gfortran I think that needs to be cleaned up. Otherwise looks good to me.

CMake/cdat_modules/esmf_pkg.cmake
chaosphere2112: I didn't personally test with 5.0 or 5.1; I can definitely state that 5.2 doesn't work, though.

CMake/cdat_modules/esmf_pkg.cmake
chaosphere2112: The previous required version was 4.3, which this would match.

aashish24:
@dnadeau4 the patch wouldn't break the older GCC would it?

aashish24:
Other than that, it looks good to me. :+1: thanks for pushing it.

CMake/cdat_modules/esmf_pkg.cmake
RCHG: I confirm that this problem was present also in Debian (in my case in the current testing version). I have changed as explained,

**CMake/cdat_modules/esmf_pkg.cmake**

From 
```bash
if(${Fortran_MAJOR_VERSION} GREATER 3 AND ${Fortran_MINOR_VERSION} GREATER 1)
```
to
```bash
if(${Fortran_MAJOR_VERSION} GREATER 4 AND ${Fortran_MINOR_VERSION} GREATER 0)
```
and the CMAKE seems to work.

CMake/cdat_modules/esmf_pkg.cmake
dnadeau4: ESMF was compiled with gfortran 5.2.1  
* ./lib0/Linux.gfortran.64.openmpi.default/libesmf.so
* ./lib0/Linux.gfortran.64.openmpi.default/libesmf.a

I changed 4.3 to 4.1 due to the test above, but I debated about keeping 4.3 in the ERROR message.  If somebody use gfortran 4.1 the test will pass.  I don't know if ESMF will compile though, the documentation says ESMF requires gfortran 4.3 for linux

crunchy: uses gfortran 4.4.7  and  I have gfortran 5.2.1,  that is why I used MINOR_VERSION  greater than 1.

* http://sourceforge.net/p/esmf/esmf/ci/master/tree/build_config/Linux.gfortran.default/

aashish24: @dnadeau4 why not keep the check for 4.3 then?

aashish24:
@dnadeau4 are you going to update it? with CMake version LESS ...of CMake..

dnadeau4:
I will when I get a chance.

dnadeau4:
@aashish24 Done!

dnadeau4:
@doutriaux1 can you take a look at the CMAKE version for gfortran.

aashish24:
thanks @dnadeau4 on a quick look at it it looks good. I will have a closer look at it.

doutriaux1:
@aashish24 since it's all CMake I will let you d the final merge, but it looks good.

aashish24:
LGTM :+1: thanks @dnadeau4

Issue #1650, opened 2015-10-30
Patterns and Opacity


chaosphere2112:
Let's continue the debate from #1637 here instead...

I'll leave this post as just stating the issues, and we can make our arguments below.

The issues at hand: the API for patterns, and the `opacity`/`fillareaopacity` attribute's behavior

## The historical behavior

Historically, there was no `opacity` attribute, so having one is a nice step forward.

The difference between a `pattern` and a `hatch` looked like this (though this is from the current version):

![pattern_vs_hatch](https://cloud.githubusercontent.com/assets/718512/10849111/6224f7a8-7edc-11e5-8c02-0f0ed5cd9d30.png)

```python
import vcs

x = vcs.init()
fa = x.createfillarea()
fa.x = [[.25, .5, .5, .25], [.5, .75, .75, .5]]
fa.y = [[.25, .25, .75, .75], [.25, .25, .75, .75]]
fa.style = ["pattern", "hatch"]
fa.color = 50
fa.index = 9

x.plot(fa)
```

A `pattern` would plot a black pattern, and a `hatch` would plot a colored pattern. That was basically the only difference.

## The current behavior

Currently, if you leave `opacity` at its default value, you get the same plot:

![pattern_vs_hatch](https://cloud.githubusercontent.com/assets/718512/10849111/6224f7a8-7edc-11e5-8c02-0f0ed5cd9d30.png)

```python
import vcs

x = vcs.init()
fa = x.createfillarea()
fa.x = [[.25, .5, .5, .25], [.5, .75, .75, .5]]
fa.y = [[.25, .25, .75, .75], [.25, .25, .75, .75]]
fa.style = ["pattern", "hatch"]
fa.color = 50
fa.index = 9

x.plot(fa)
```

If you set `opacity` on a hatch, you get something like this:

![hatch_opacity](https://cloud.githubusercontent.com/assets/718512/10849877/14cf58e6-7ee0-11e5-8810-06a53c9f5cf7.png)

```python
import vcs

x = vcs.init()
fa = x.createfillarea()
# Draw two hatches, one transparent, with the transparent one overlaid on a solid one
fa.x = [[.25, .5, .5, .25], [.5, .75, .75, .5], [.5, .75, .75, .5]]
fa.y = [[.25, .25, .75, .75], [.25, .25, .75, .75], [.25, .25, .75, .75]]
fa.style = ["hatch", "hatch", "solid"]
fa.opacity = [100, 25, 100]
[...]

PR #1651, opened 2015-10-30
Improved pattern size

chaosphere2112:
When I was doing the test scripts in #1650, I saw that the patterns looked really wonky on large fillareas. Turns out, it was guessing that they were 32x64 pixels, which was, uh, incorrect. Made it so we can pass in screen size to the `fillareautils` code to get more realistic sizes, updated baselines to match. I also added a test for larger fillareas that makes sure they look fine.


doutriaux1:
hum weird failure on annie, VTK issue?

doutriaux1:
Actually all Linux are offset from Macs. @sankhesh @dlonie @aashish24 could that be VTK?

Packages/vcs/Lib/vcsvtk/fillareautils.py
```diff
@@ -33,8 +34,14 @@ def make_patterned_polydata(inputContours, fillareastyle=None,
...
-    xres = int(4.0*xBounds)
-    yres = int(4.0*yBounds)
+    if xBounds <= 1 and yBounds <= 1 and size is not None:
+        xBounds *= size[0]
+        yBounds *= size[1]
+        xres, yres = int(xBounds), int(yBounds)
+    else:
+        xres = int(4.0*xBounds)
+        yres = int(4.0*yBounds)
+    print xres, yres, size
```
sankhesh: Debug statement left over?

Packages/vcs/Lib/vcsvtk/fillareautils.py
```diff
@@ -33,8 +34,14 @@ def make_patterned_polydata(inputContours, fillareastyle=None,
     # Scaled the size to 2 times to make the pattern image of a finer resolution
     xBounds = bounds[1] - bounds[0]
     yBounds = bounds[3] - bounds[2]
-    xres = int(4.0*xBounds)
-    yres = int(4.0*yBounds)
+    if xBounds <= 1 and yBounds <= 1 and size is not None:
+        xBounds *= size[0]
```
sankhesh: Not sure if I understand this logic.
IMHO, the parameters `(xres, yres)` should be dependent on the bounds of the polydata and not on the screen size.

Wouldn't this produce different results for different displays?

Packages/vcs/Lib/vcsvtk/fillareautils.py
chaosphere2112: Yup, I'll remove.

Packages/vcs/Lib/vcsvtk/fillareautils.py
chaosphere2112: Here's the reason for this pull request:

```python
import vcs
fa = vcs.createfillarea()
fa.x = [0, 0, 1, 1]
fa.y = [0, 1, 1, 0]
fa.style = "hatch"
fa.color = 242
fa.index = 5
canvas = vcs.init()
canvas.plot(fa)
```

Using master, you get this:
busted

Using this branch, you get this:
size_adjust

For fillareas, the bounds are always between 0 and 1; this means we wind up applying a 16x32 pixel pattern to a *giant* fillarea. If the texture were being applied in a tiling manner, this wouldn't be an issue, but since we create one texture map for everything, this is going to look really terrible.

Also, no, I don't think it will create different results for different displays. The size passed in is from `vtkRenderWindow.GetSize()`, so I think it should always generate the same output for a given render window size.

Also, I tested with multiple resolutions and it appears to have always been the same. That said, I should make it so the test specifies a window size.

chaosphere2112: Oh, nevermind, don't need to change the test, it's already set correctly.

sankhesh: :stuck_out_tongue:  I see the issue.

But, I thought one of @doutriaux1's recent changes makes render window size dependent on screen size.

chaosphere2112: It does, but the pattern that is output by this will be the same for any given render window size. While the default size may differ, the `png` function will resize the window (triggering a rebuild of this pattern) and the output should match up.

sankhesh: Sorry @chaosphere2112 that still does not sound right. I would not expect my view to change when taking a screenshot. Say, I set up the view which shows 20 pattern lines and take a screenshot that would resize the render window to half its size (say), then by this logic, I would get 10 pattern lines in the same area which is not a faithful reproduction of the scene. Am I right?

doutriaux1: @sankhesh is kind of right you want your pattenr to be somewhat predictable, can we try to scale this pattern by the size of the window or something?

@chaosphere2112 with your changes would a fillarea pattern be different based on different values of the fa.worldcoordinate ?

sankhesh: @chaosphere2112 Perhaps you could use the Normalized Viewport coordinates instead of render window size?

chaosphere2112: @doutriaux1 It is related to render window size currently, yes.

@sankhesh Ah, I think I understand our problem (and the solution). `fillarea`s are being set up with normalized viewport coordinates, not viewport coordinates. I'm going to see what I can do about fixing that issue; then the code here should work properly (as the bounds will be greater than 1).

aashish24:
@chaosphere2112 @sankhesh ..ping!

chaosphere2112:
I've been working on this nonstop for the last week. I think I've just about got it squared away, just fixing a small issue where isofill pattern sizes and fillarea pattern sizes aren't appearing to be the same (I am pretty sure this is a camera thing).

chaosphere2112:
@doutriaux1 @aashish24 @sankhesh OK, this is ready for review finally.

doutriaux1:
not sure what is going on on crunchy...

doutriaux1:
@aashish24 it looks good to me. It passes the tests too. Just not sure why it's trying to put @jbeezley code back in. Did it get removed somehow? Anyhow if you're ok with it please merge.

jbeezley:
That's a github issue.  It doesn't detect the duplicate commits resulting from a merge.  I often just rebase onto master rather than merge.

chaosphere2112:
@jbeezley Yeah, I did some bad things to my local git repo with this PR. I think I split a branch off of this branch, rebased off master, rebased again off master, then merged the rebased branch back into the original? There were hijinks. They ensued.

Packages/vcs/Lib/vcsvtk/fillareautils.py
```diff
@@ -33,19 +38,18 @@ def make_patterned_polydata(inputContours, fillareastyle=None,
     # Scaled the size to 2 times to make the pattern image of a finer resolution
```
sankhesh: This comment is no longer true.

sankhesh:
:+1: LGTM except a minor nitpick

aashish24:
@chaosphere2112 why don't you rebase again? It will make a nice clean branch..

chaosphere2112:
@aashish24 The history is pretty mutilated with this branch. I'm going to cherry pick out the commits into a new branch with a clean history and resubmit.

PR #1657, opened 2015-11-02
Do not wrap again in update_input

### LINKED ISSUE #1086: animation does not work on projected plots
```python
import cdms2,vcs,os

f=cdms2.open(os.path.join(vcs.sample_data,"clt.nc"))

print f.variables.keys()

ice = f("clt",latitude=(90,30))
print ice.shape

x=vcs.init()

gm = x.createisofill()

levs = vcs.mkscale(ice.min(),ice.max())
cols=vcs.getcolors(levs)

gm.levels= levs
gm.fillareacolors = cols
gm.projection="polar"
x.plot(ice,gm)

x.animate.create()

import time
time.sleep(1)
print "Creating"

x.animate.save("ice.mp4")

print x.close(preserve_pngs=True)
```

sankhesh:
This change fixes the issue wherein no output was being produced
when animating projected data. Each draw_frame call invokes
VTKPlots::update_input which in turn calls doWrap on the
actor. Trying to wrap projected data fails.

Fixes #1086

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1294,6 +1294,10 @@ class VTKVCSBackend(object):
                 vg.GetPointData().AddArray(w)
                 ports[0].SetInputData(vg)
 
+            projection = None
+            if "vtk_backend_geo" in vtkobjects:
+                projection = vtkobjects["vtk_backend_geo"]
+
             if "vtk_backend_actors" in vtkobjects:
                 i = 0
                 for a in vtkobjects["vtk_backend_actors"]:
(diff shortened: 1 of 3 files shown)


sankhesh:
@doutriaux1 @aashish24 Please review

aashish24:
@sankhesh this is great! Makes sense and LGTM :+1: Should we add a test for projected dataset animation as well?

sankhesh:
Yup. I can add one.

aashish24:
@sankhesh can add one soon so that we can make it done?

aashish24:
@sankhesh we should really get this one in master asap. Can you please add test soon?

Thanks,

sankhesh:
Baselines added in UV-CDAT/uvcdat-testdata#81

sankhesh:
@doutriaux1 While adding the test for this issue, I found that meshfill plots with mercator projection fails. Documented here: #1671

sankhesh:
Not sure why we have those test failures. They don't fail on my machine. 
@aashish24 @doutriaux1  Can you reproduce those failures?

aashish24:
@sankhesh which machine you are seeing these failures?

sankhesh:
annie, crunchy

doutriaux1:
@sankhesh both annie and crunchy are running behind Xvfb. Can you try with Xvfb?

aashish24:
@sankhesh any luck with testing?

sankhesh:
@aashish24 I've pushed a new change to test only the last image of the animation as you suggested. Waiting on the dashboards to complete.

aashish24:
thanks @sankhesh much appreciated!

PR #1660, opened 2015-11-04
Stop configuring if test data fails to download

jbeezley:
This is mostly just to see what is going on with the dashboard machines, but we really should treat this as a fatal error.


aashish24:
+1 for this. yes, it should be an error.

PR #1661, opened 2015-11-04
Use cmake to download uvcmetrics test data

jbeezley:
Replace the old python script with a pure cmake implementation.  Also, the download will now occur build time rather than configure time.  This invalidates #1660.


doutriaux1:
Nice! I've been meaning to do this for ever now!

doutriaux1:
All good. Merging.

aashish24:
:clap:

PR #1667, opened 2015-11-09
Window flickering when png

doutriaux1:
@aashish24 @chaosphere2112 @sankhesh

### DIFF
--- a/Packages/vcs/Lib/Canvas.py
+++ b/Packages/vcs/Lib/Canvas.py
@@ -4747,7 +4747,6 @@ Options:::
 
         W, H = self._compute_width_height(
             width, height, units)
-        print "Requesting png with:",width,height
         return self.backend.png(
             file, W, H, units, draw_white_background, **args)
 
@@ -4766,9 +4765,9 @@ Options:::
  Example of Use:
     a=vcs.init()
     a.plot(array)
-    a.png('example')       # Overwrite a postscript file
-    a.png('example', width=11.5, height= 8.5)  # US Legal
-    a.png('example', width=21, height=29.7, units='cm')  # A4
+    a.pdf('example')       # Overwrite a postscript file
+    a.pdf('example', width=11.5, height= 8.5)  # US Legal
+    a.pdf('example', width=21, height=29.7, units='cm')  # A4
 """
         if units not in [
                 'inches', 'in', 'cm', 'mm', 'pixel', 'pixels', 'dot', 'dots']:
@@ -4914,7 +4913,7 @@ Options:::
         """Is the Canvas opened?"""
         return self.backend.isopened()
 
-    def _compute_width_height(self, width, height, units, ps=True):
+    def _compute_width_height(self, width, height, units, ps=False):
         dpi = 72.  # dot per inches
         if units in ["in", "inches"]:
             factor = 1.
@@ -4984,6 +4983,8 @@ Options:::
             tmp = W
             W = H
             H = tmp
+        if H % 2 == 1:
+            H-=1
         return W, H
 
     def postscript(self, file, mode='r', orientation=None, width=None, height=None,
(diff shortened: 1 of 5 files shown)


doutriaux1:
goes with: https://github.com/UV-CDAT/uvcdat-testdata/pull/79

doutriaux1:
@aashish24 @jbeezley @sankhesh I'm not sure what happened on the macbot1 ?

Packages/vcs/Lib/Canvas.py
```diff
@@ -4912,7 +4913,7 @@ def isopened(self):
         """Is the Canvas opened?"""
         return self.backend.isopened()
 
-    def _compute_width_height(self, width, height, units, ps=True):
+    def _compute_width_height(self, width, height, units, ps=False):
```
aashish24: Why false now?

Packages/vcs/Lib/Canvas.py
```diff
@@ -5018,7 +5021,7 @@ def postscript(self, file, mode='r', orientation=None, width=None, height=None,
 
         # figures out width/height
         W, H = self._compute_width_height(
-            width, height, units)
+            width, height, units, ps=True)
```
aashish24: I see. so we treat ps as a special case..

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -422,8 +424,13 @@ def portrait(self, W=-99, H=-99, x=-99, y=-99, clear=0):
     def landscape(self, W=-99, H=-99, x=-99, y=-99, clear=0):
         self.resize_or_rotate_window(W, H, x, y, clear)
 
-    def initialSize(self):
+    def initialSize(self, width=None, height=None):
         # Gets user physical screen dimensions
+        if isinstance(width,int) and isinstance(height,int):
```
aashish24: space between width, and int please :cry:

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -458,6 +465,7 @@ def isopened(self):
 
     def geometry(self, x, y, *args):
         self.renWin.SetSize(x, y)
+        self._lastSize = (x, y)
```
aashish24: why we are storing lastSize?

aashish24: I know in the past there was an issue with vistrails. Is that still true? I would rather prefer not to have this convoluted logic behind previous and the current size if possible.

Packages/vcs/Lib/Canvas.py
doutriaux1: This was only for postscript.

Packages/vcs/Lib/Canvas.py
doutriaux1: Yes

doutriaux1: That was actually the reason for the flickering, it was going through that bit and changed the size

Packages/vcs/Lib/VTKPlots.py
doutriaux1: Yes I always forget to run flake8, will push changes for flake8 soon.

Packages/vcs/Lib/VTKPlots.py
doutriaux1: I'm not sure but since we do it everywhere else where the size changes I added it as well. @chaosphere2112 do we still need this when we switch to your GUI

doutriaux1:
@aashish24 ping

aashish24:
@doutriaux1 changes looks good to me :+1: we can clean up lastSize stuff afterwards (this was anyways exists currently).

doutriaux1:
great! Feel free to merge :wink:

aashish24:
LGTM merging

PR #1692, opened 2015-11-14
fix configure arguments separator

dnadeau4:
+set(readline_conf_args "--with-curses\;--disable-static\;--enable-shared")


CMake/cdat_modules/readline_external.cmake
```diff
@@ -2,7 +2,7 @@ set(readline_source "${CMAKE_CURRENT_BINARY_DIR}/build/readline")
 set(readline_install "${cdat_EXTERNALS}")
 set(readline_conf_args)
 
-set(readline_conf_args "--with-curses --disable-static --enable-shared")
+set(readline_conf_args "--with-curses\;--disable-static\;--enable-shared")
```
aashish24: you shouldn't need to do use backslash here

dnadeau4: When I compile with a "space" cmake fails on my Ubuntu machine at home.  I tried "^^" and it failed also, I will take away the backspace, but we should be consistent.  in ffmpeg_external.cmake the arguments are set to "^^".

PR #1696, opened 2015-11-18
Now will work if there is no .uvcdat directory

chaosphere2112:
Fixes a bug where the `uvcdat` launcher script would crash if there was no .uvcdat directory (or if the user specified a path pointing at a directory that doesn't exist for output).


aashish24:
LGTM :+1: will wait for Mac bots

PR #1699, opened 2015-11-19
fixed issue where old scr colormap file would not load opacity

doutriaux1:
@aashish24 @sankhesh @chaosphere2112 just found a tiny bug. Please merge.

### DIFF
--- a/Packages/vcs/Lib/colormap.py
+++ b/Packages/vcs/Lib/colormap.py
@@ -33,7 +33,7 @@ def process_src(nm, code):
     for i in range(255):
         val = numbers[i * 3:i * 3 + 3]
         if val != ():
-            d[i] = list(val)+[100]
+            d[i] = list(val) + [100]
     cp = Cp(nm)
     cp.index.data.update(d)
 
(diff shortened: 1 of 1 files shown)


Packages/vcs/Lib/colormap.py
```diff
@@ -33,7 +33,7 @@ def process_src(nm, code):
     for i in range(255):
         val = numbers[i * 3:i * 3 + 3]
         if val != ():
-            d[i] = list(val)
+            d[i] = list(val)+[100]
```
aashish24: @doutriaux1 can you exaplain bit more what this code is doing?

doutriaux1: old colormaps only had r,g,b values so when loading back in it only stores 3 values, then later on when using getcolor we unpack it into r,g,b,a causing a failure. Voila.

sankhesh: Not Flake8 compliant.

Need spaces around binary operator "+".

doutriaux1: yet flake8 passes, bwwwaaaaaa I beat flake8!!!!!

aashish24:
LGTM :+1:

Issue #1645, opened 2015-10-28
Isofill Blank Spots on Solid with Indices/Levels


chaosphere2112:
@aashish24 @sankhesh @doutriaux1 Just ran into this one while testing my performance branch, then went back and found it on master.

```python
import vcs, cdms2
x = vcs.init()
f = cdms2.open(vcs.sample_data + "/clt.nc")
s = f("clt")

iso = x.createisofill()
iso.levels = range(0, 100, 10)
iso.fillareaindices = range(1, 11)
iso.ext_2 = True

x.plot(s, iso)
```
 
will give you this:

![out](https://cloud.githubusercontent.com/assets/718512/10796902/6b696c16-7d5d-11e5-9578-fbd053442c3e.png)

Obviously, `fillareaindices` shouldn't have any effect on a `solid` isofill, so that's a pretty weird behavior.

chaosphere2112:
The `ext_2` is insignificant, but was in the script that I used to generate the image.

chaosphere2112:
Also with patterns:

```python
import vcs, cdms2
x = vcs.init()
f = cdms2.open(vcs.sample_data + "/clt.nc")
s = f("clt")

iso = x.createisofill()
iso.levels = range(0, 100, 10)
iso.fillareastyle = "pattern"
iso.fillareaindices = range(1, 11)
iso.fillareacolors = range(20, 220, 20)
iso.fillareaopacity = [100] * 10

x.plot(s, iso)
```

![pattern_white_tri](https://cloud.githubusercontent.com/assets/718512/10798686/1f375d9a-7d66-11e5-8ee5-3e9b5189cc50.png)

sankhesh:
@danlipsa Could you please look into this next?

aashish24:
@danlipsa I am hoping this would a tiny bug somewhere. @doutriaux1 we may have to patch our release.

danlipsa:
Seems the problem is how vtkBandedPolyDataContourFilter fills a band. Attached are two pictures: first isocontours at 40 and 50 then a band for 40-50.
![contours-40-50](https://cloud.githubusercontent.com/assets/6180727/11284386/6d8f02f4-8ed7-11e5-8017-8fa1799f27d0.png)
![contours-filled-40-50](https://cloud.githubusercontent.com/assets/6180727/11284387/6d9283a2-8ed7-11e5-95e3-acaa80f3e4d2.png)

PR #1700, opened 2015-11-19
BUG #1645: Blank spots on solid fillareastyle with fillareaindices

danlipsa:
The solid fillareastyle was treated as a pattern style in that
each level was extracted and rendered individually. Because of a
bug in vtkBandedPolyDataContourFilter lower isocontours differ
from upper isocontours which results in blank spots. In this fix
we trigger merging of levels which fixes the problem for solid fill.
We'll fix the problem for pattern fill in a subsequent commit.

### DIFF
--- a/Packages/vcs/Lib/vcsvtk/isofillpipeline.py
+++ b/Packages/vcs/Lib/vcsvtk/isofillpipeline.py
@@ -102,6 +102,7 @@ class IsofillPipeline(Pipeline2D):
         x1, x2, y1, y2 = vcs.utils.getworldcoordinates(self._gm,
                                                        self._data1.getAxis(-1),
                                                        self._data1.getAxis(-2))
+        assert(style != 'solid' or len(tmpLevels) == 1)
         for i, l in enumerate(tmpLevels):
             # Ok here we are trying to group together levels can be, a join
             # will happen if: next set of levels continues where one left off
@@ -123,11 +124,11 @@ class IsofillPipeline(Pipeline2D):
             for j, color in enumerate(tmpColors[i]):
                 r, g, b, a = self.getColorIndexOrRGBA(_colorMap, color)
                 if style == 'solid':
-                    tmpOpacity = tmpOpacities[i]
+                    tmpOpacity = tmpOpacities[j]
                     if tmpOpacity is None:
                         tmpOpacity = a / 100.
                     else:
-                        tmpOpacity = tmpOpacities[i] / 100.
+                        tmpOpacity = tmpOpacities[j] / 100.
                     lut.SetTableValue(j, r / 100., g / 100., b / 100., tmpOpacity)
                 else:
                     lut.SetTableValue(j, 1., 1., 1., 0.)
--- a/Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
+++ b/Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
@@ -346,7 +346,7 @@ class BoxfillPipeline(Pipeline2D):
         geos = []
         wholeDataMin, wholeDataMax = vcs.minmax(self._originalData1)
         _colorMap = self.getColorMap()
-
+        assert(style != 'solid' or len(tmpLevels) == 1)
         for i, l in enumerate(tmpLevels):
             # Ok here we are trying to group together levels can be, a join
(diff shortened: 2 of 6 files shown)


danlipsa:
@sankhesh @aashish24 @doutriaux1 Please review.

Packages/vcs/Lib/vcsvtk/isofillpipeline.py
```diff
@@ -93,9 +93,11 @@ def _plotInternal(self):
         tmpColors = []
         tmpIndices = []
         tmpOpacities = []
-        indices = self._gm.fillareaindices
-        opacities = self._gm.fillareaopacity
         style = self._gm.fillareastyle
+        # levels would not be merged if the user sets fillareaindices for a 
+        # 'solid' style resulting in a less efficient plotting.
+        indices = (None if (style == "solid") else self._gm.fillareaindices)
+        opacities = self._gm.fillareaopacity
```
aashish24: LGTM but @sankhesh should approve it.

sankhesh:
@danlipsa I would say this is a workaround and the real issue is in vtkBandedPolyDataContourFilter. Fixing the filter would fix this issue for all fill styles. 

**Question:** If the above hypothesis is correct, do we need this workaround? 
Running through vtkBandedPolyDataContourFilter once and adding just one renderer to the scene for solid fills would be better in terms of performance and picking.

Moreover, this issue will present even if there is just one index and multiple colors or opacities.

FYI, this code needs to be rebased on master. This calculation was moved to the superclass.

doutriaux1:
I think @sankhesh is right, beside @chaosphere2112 pointed out that it doesn't work either for hatches and patterns (although it is harder to tell)

aashish24:
> I think @sankhesh is right, beside @chaosphere2112 pointed out that it doesn't work either for hatches and patterns (although it is harder to tell)

@danlipsa pointed out that this is only for solid fills. I believe @danlipsa is planning to push a separate branch for patterns.

danlipsa:
Indeed, fixing vtkBandedPolyDataContourFilter would fix things for solid fill as well. I think this patch is still worth doing as without it, we do more work than required. There is no need to split the dataset in bands and render each of them individually (for solid fill). We can just render one band and set colors and opacities in the LUT.

chaosphere2112:
Yeah, I agree; it would be better to not actually bother running through the pattern code with `solid` fills.

doutriaux1:
Broken tests. @chaosphere2112 I will let you take the lead on this and merge if you are OK when tests pass. Thx

danlipsa:
@doutriaux1 is this good to go?

danlipsa:
@sankhesh pointed out that I should make the same changes to meshfillpipeline. This new pull request adds this and also adds new tests baselines for meshfill:
https://github.com/UV-CDAT/uvcdat-testdata/pull/87

chaosphere2112:
@danlipsa I'll test this today and hopefully merge.

aashish24:
@danlipsa were you able to fix the legend?

danlipsa:
@aashish24 yes, the problem was with opacities rather than the legend/colors.

aashish24:
Ah thanks @danlipsa so this will fix the blanking for solids. I am assuming that we still need fix for patterns?

danlipsa:
@aashish24 yes, we still need the fix for patterns however, that bug is not easy to spot. See the following image:
spots_patterns

danlipsa:
@sankhesh @aashish24 @chaosphere2112 Please review. It seems that the relevant tests pass.

sankhesh:
@danlipsa You might want to increase the image comparison threshold for meshfill tests. See https://open.cdash.org/viewTest.php?onlyfailed&buildid=4115057

danlipsa:
See the VTK bug report for vtkBandedDataContourFilter. Will said he is going to look at it this week or next.
http://www.vtk.org/Bug/view.php?id=15845

chaosphere2112:
I got 100% passed tests, and looking at my test case from #1645, it seems to be fixed. Changes seem straightforward. I'll merge!

chaosphere2112:
Oh, whoops; probably should have waited till the thresholds got bumped up for the meshfill tests.

chaosphere2112:
... or maybe that did get in. Don't mind me, I'm quite insane, you know.

PR #1703, opened 2015-11-20
fixed tag point for external repos

doutriaux1:


doutriaux1:
@aashish24 @sankhesh @chaosphere2112 when we are ready. I already tagged all externals repos as well so if we change anything in libcdms, cmor, vtk, vistrails or uvcmetrics that we want in the release we will need to move their `uvcdat-2.4.0` tag point as well

durack1:
@doutriaux1 paraview is out?

doutriaux1:
I must have missed a tag point on some repo here...

aashish24:
> @doutriaux1 paraview is not included in the build anymore? Doesn't it also need a tag point set?

@durack1 ParaView is out. We no longer build ParaView.

Issue #1106, opened 2015-03-03
Continents are off


doutriaux1:
When plotting data with a user we noticed the continents seem to be a bit to much up north.
1.5.1 is ok. Png to follow soon

doutriaux1:
our version (BAD)
![bad](https://cloud.githubusercontent.com/assets/2781425/6473566/0d72ca5e-c1af-11e4-9f5e-ccb18af166eb.png)

1.5.1 version (good)
![good](https://cloud.githubusercontent.com/assets/2781425/6473577/1e35cab2-c1af-11e4-895d-54e4bc164432.png)

doutriaux1:
actually the continents look good, it's almost as if the data is shifted rather than continents

doutriaux1:
@dlonie I think it is because we set the data on "cellData" rather than "pointData" I'm using ployDataMApper and DatasetMAppers, do they do anything with the grid if data is on cells? It is odd.

Please use the same script/data (bad.nc) I sent earlier and switch between isofill or boxfill for gms, you will see the data shift dramatically when switching. Moving isoline/fills from cellData to pointData seems to have fixed this.

Any idea?

Bellow the good (isofill) and bad/shifted (boxfill) data
good:
![ok](https://cloud.githubusercontent.com/assets/2781425/6518139/c9757ed2-c357-11e4-9bea-1a807813c276.png)
and bad:
![bad](https://cloud.githubusercontent.com/assets/2781425/6518142/d7b1b042-c357-11e4-99d0-7e2aefd10996.png)

doutriaux1:
and good version of boxfill, generated with 1.5.1 (excpet for colors)
![bad](https://cloud.githubusercontent.com/assets/2781425/6518283/6569bac8-c359-11e4-9da1-2c628f673895.png)

doutriaux1:
@dlonie any ideas?

dlonie:
I suspect it has something to do with the transforms that are used to
resize / translate the data (the fitToViewport stuff, IIRC). I'd start
looking there.

It seems to me that it'd be more robust to not rescale/stretch/etc the data
with a transform, but rather leave the data in the same coordinate system
and adjust the viewport/camera to get the desired view -- It's fewer things
to keep track of and less error prone. Is there a reason for using the
transforms that I'm missing?

doutriaux1:
@aashish24 you designed this at the very beginning of the VTK rewrite, can you take a look please? 
[...]

Issue #1462, opened 2015-07-21
Bad rendering of aeqd projection


sankhesh:
The following image is one of the baselines in the vcs testing suite (for `test_vcs_basic_boxfill_aeqd_proj`)

![test_vcs_basic_boxfill_aeqd_proj](https://cloud.githubusercontent.com/assets/936354/8802994/6a0ae9a8-2f92-11e5-9251-e521ef2bd5d2.png)

There is a visible seam in the center of the image. The blob over Sahara in Africa is just half shown purple.

dlonie:
The new refactoring branch fixes a number of similar issues. See the meshfill-related baselines in UV-CDAT/uvcdat-testdata#58. The main issue there was that `doWrap` uses the scale factor set by `fitToViewport`, but was called before `fitToViewport`, so it was wrapping using the unscaled bounds.

Not sure if this is the same issue, but it sounds similar. My fix in that branch is tied up with the new vtk python filters that segfault when they are leaked, so we'll need to address the leaks before that can be tried. But it might help for tracking this down.

Again, it would be a good idea (in my opinion) to have a user/scientist/vcs guru go through the baselines and the tests to identify which ones have issues. Seems like we're finding invalid "reference" images fairly regularly, and it would go a long way towards improving the quality of UVCDAT to ensure that our reference images are actually correct. Just my 2c.

doutriaux1:
northern europe continents are all messed up as well. I think this issue has been present since before the refactoring though. I guess we can double check with the baseline history.

durack1:
@dlonie are there plans to add some projection examples (Mollweide, Robinson) into the test suite? These are where some of the more obvious errors have appeared in `2.2.0`

dlonie:
@durack1 I'm not working on VCS any longer, only on VTK, so I have no plans to add further tests for VCS. Please talk with @sankhesh or @doutriaux1 about VCS specific issues.

doutriaux1:
@sankhesh is that in your area of expertise? @aashish24 should we find someone at Kitware who is an expert at these? Lots of work is needed with projections it might be worth it.

aashish24:
@doutriaux1 is this a projection issue or a wrapping issue?

doutriaux1:
The latest I've seen are wrapping issues, drawing the boxes/isofills accros the globe when going 5->355, instead of splitting it into 2 boxes. One way to "know" how to treat cells like this is to look if the vertices go clockwise vs counterclockwise.

[...]

PR #1706, opened 2015-11-24
BUG #1106: wrong dataset created from nc file with decreasing lat and…

danlipsa:
… bounds

We did not handle correctly a nc file which specifies bounds and
has decreasing latitude or longitude.


danlipsa:
@sankhesh @aashish24 @doutriaux1 Please review.

aashish24:
@danlipsa this is great! I will try it later unless someone else beats me to it. Do we have a test?

danlipsa:
@aashish24 No, I did not add one. Should I add just the script/data that was used to show the bug?

danlipsa:
@doutriaux1 @durack1 There are some tests failing where the test image seems correct and baseline seems wrong. Could you confirm that, for instance by looking at old baselines? See the boxfill tests in https://open.cdash.org/viewTest.php?onlyfailed&buildid=4116767

durack1:
@danlipsa I see what you mean but I'm not certain what changes have occurred that could have caused this - maybe @chaosphere2112 can answer if @doutriaux1 is not currently available..? I wonder if some of the baselines need to be updated?

sankhesh:
I'm curious as to why the isofill tests are not failing.

chaosphere2112:
@danlipsa @sankhesh @durack1 I'm waiting on the internet to stop being terrible here, then I'll take a quick look

chaosphere2112:
@danlipsa Looks like we still need to bump up the meshfill tests' threshold a bit more (they're failing on Annie; I'd say go to 40). As for the actual content of this PR, I tend to agree that the updated images look better. I'd want to wait on @doutriaux1 to comment when he gets back before merging this one, though.

aashish24:
@danlipsa can you look into these failing tests? I see that box fills are failing as well

danlipsa:
@aashish24 see the discussion earlier in the pull request. We believe the test images are correct and the baseline are wrong. We wait for @doutriaux1 to take a look at this as well.

aashish24:
@doutriaux1 can you look at this branch and merge it if all looks good?

doutriaux1:
will retrigger now and take a look tomororw morning

aashish24:
@danlipsa if @doutriaux1 approves / agrees that new image (test image) is better than I would suggest you or @doutriaux1 update the baselines specifically for the _NH_ ones..

doutriaux1:
@danlipsa building now.

doutriaux1:
@danlipsa did you run this test on the seaice test? I think it should fix it. In the mean time I will create a baseline update. Also if it fixes the sea-ice could you add it as a test case? (put the nc file in uvcdat-testdata repo under data )

danlipsa:
@doutriaux1 yes, it fixes the sea ice test. @aashish24 also suggested adding a test for the sea ice and I meant to answer him: My thought is that we have all these other tests that exhibited the same faulty behavior - an empty area at the pole and all data shifted. Updating the baseline for these tests would be enough to test that we maintain the correct behavior. Adding another test for sea ice would not test anything new. What do you guys think?

doutriaux1:
@danlipsa the other tests while exhibiting this issue were not meant to test this though, so I think it's worth a dedicated test.

doutriaux1:
on that note @aashish24 the projection is still wrong....

aashish24:
> on that note @aashish24 the projection is still wrong....

@doutriaux1 are you talking about aeqd projection?

doutriaux1:
updated baselines at: https://github.com/UV-CDAT/uvcdat-testdata/pull/89

doutriaux1:
@danlipsa when you added the test if these pass happily then let's merge.

doutriaux1:
yes look at the baseline they are "cut" at the greenwich meridian, but that's another `known`ish issue.

danlipsa:
@doutriaux1 @aashish24 OK, I'll add a test for the sea ice.

aashish24:
+1 @danlipsa

sankhesh:
@doutriaux1 The aeqd projection rendering issue is #1462

sankhesh:
One thing that is strange to me is why did none of the `isofill` tests fail with this change? The baselines were just updated for `boxfill` tests.

doutriaux1:
@sankhesh I think isofill use pointData rather than cellData.

sankhesh:
Yes, but this change does not have anything to do with the representation. As far as I can tell, it just changes the data extents.

PR #1709, opened 2015-11-25
Fix initial size

chaosphere2112:
Added a test for initial sizing of window (you can pass a `geometry` kwarg to vcs.init to specify window dimensions), and fixed some issues with the logic. Also tests to make sure `open(width=X, height=X)` works properly.

### DIFF
--- a/Packages/vcs/Lib/Canvas.py
+++ b/Packages/vcs/Lib/Canvas.py
@@ -946,11 +946,12 @@ class Canvas(object):
                 height = geometry.get("height", None)
 
                 check_vals = [v for v in (width, height) if v is not None]
-                VCS_validation_functions.checkListOfNumbers(self, 'geometry', check_vals, 
+                VCS_validation_functions.checkListOfNumbers(self, 'geometry', check_vals,
                                                             minvalue=1, minelements=1, maxelements=2, ints=True)
             elif type(geometry) in (list, tuple):
                 width, height = VCS_validation_functions.checkListOfNumbers(self, 'geometry', geometry,
-                                                                            minvalue=1, minelements=2, maxelements=2, ints=True)
+                                                                            minvalue=1, minelements=2,
+                                                                            maxelements=2, ints=True)
             else:
                 raise ValueError("geometry should be list, tuple, or dict")
             geometry = {"width": width, "height": height}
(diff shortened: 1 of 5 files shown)


aashish24:
@chaosphere2112 flake8_vcs is failing..

aashish24:
@chaosphere2112 @doutriaux1 lot of tests are failing on Linux-RH6. Looking into it.

aashish24:
@doutriaux1 could it be because of VTK SHA (wrong one)

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -328,8 +328,10 @@ def createRenWin(self, *args, **kargs):
             else:
                 width = None
                 height = None
-            width = kargs.get("width", width)
-            height = kargs.get("height", height)
+            if "width" in kargs and kargs["width"] is not None:
+                width = kargs["width"]
+            if "height" in kargs and kargs["height"] is not None:
+                height = kargs["height"]
```
doutriaux1: that's essentially what the old code was doing, why the more complicated lines?

Packages/vcs/Lib/Canvas.py
```diff
@@ -860,7 +860,7 @@ def objecthelp(self, *arg):
     #                                                                           #
     ##########################################################################
     def __init__(self, gui=0, mode=1, pause_time=0,
-                 call_from_gui=0, size=None, backend="vtk", geometry=None):
+                 call_from_gui=0, size=None, backend="vtk", geometry=None, bg=None):
```
doutriaux1: is that bg used anywhere?

Packages/vcs/Lib/VTKPlots.py
chaosphere2112: Because `width` and `height` are always passed into this function by the [caller](https://github.com/UV-CDAT/uvcdat/blob/8dd1c5fa0796884bb6a6c972b029761e87c21a60/Packages/vcs/Lib/VTKPlots.py#L462-L463), though most of the time they are `None`.

Packages/vcs/Lib/Canvas.py
chaosphere2112: `bg` makes it so you can call `open` without displaying the canvas. I use it in the test I added (down at the bottom), to make it so this test doesn't flash on the screen (which I know you hate :wink:). I believe I can also use it to fully background-ize a couple of the other tests I've added that flash on the screen.

Packages/vcs/Lib/VTKPlots.py
doutriaux1: I see, thx

PR #1712, opened 2015-11-30
Fixed some bugs with continents line

chaosphere2112:
* Colors weren't being properly normalized; this has been fixed.
* Alpha is now respected
* Made error handling more robust for weird input (if someone passes in a float, it'll get converted into an int then used as an index into a colormap instead of blowing up).


aashish24:
Looks correct to me..

aashish24:
LGTM :+1:

PR #1713, opened 2015-12-01
Fixed two bugs with matplotlib colormap import

chaosphere2112:
1. The `matplotlib` package doesn't have a direct import of `matplotlib.cm`, so you have to manually import the module
2. matplotlib colormaps store their values between 0 and 1, not 0 and 255.


chaosphere2112:
Baselines: UV-CDAT/uvcdat-testdata#88

Issue #1012, opened 2015-02-05
Building on OS X 10.10.2


jbeezley:
@aashish24 I think you said you were looking into this.  I found a particularly nasty problem you should be aware of.  I found that my builds have been generating thousands of crash reports like [this](https://gist.github.com/anonymous/ed7a64733073ce1be359).  

First, it looks like the `@cdat_osx_ld_flags@` shouldn't be [here](https://github.com/UV-CDAT/uvcdat/blob/master/CMake/cdat_modules_extra/cdat_common_environment.cmake.in#L2), but I don't think that is breaking anything.
However, `xcode-select` is crashing because our environment variables are loading an incompatible `libjpeg`.  If you use system `libjpg`, then it crashes like this:
```
dyld: Symbol not found: __cg_png_create_info_struct
  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
  Expected in: /Users/jbeezley/build/uvcdat/b7aada3_mpi/install/Externals/lib/libPng.dylib
 in /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
Trace/BPT trap: 5
```

I will note that this is showing up as various configure problems.  For example. #1004 is likely caused by this (in my `CURL/config.log`):
```
configure:26438: ./conftest
dyld: Symbol not found: __cg_png_create_info_struct
  Referenced from: /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
  Expected in: /Users/jbeezley/build/uvcdat/b7aada3_mpi/install/Externals/lib/libPng.dylib
 in /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
configure: line 2183: 73488 Trace/BPT trap: 5       ./conftest$ac_exeext
configure:26438: $? = 133
configure: program exited with status 133
```

As far as I can remember, this seems to be a recent problem.  I was building fine on 10.10 a couple of weeks ago.  Perhaps the system libraries were changed in the latest update.  I'm attempting to rebuild with jpg and png disabled.  I'll see how far down the rabbit hole this takes me.

jbeezley:
I just got a successful default build by using system `libjpeg` and `libpng`.  Probably should make that a default when 10.10 is detected.

potter2:
I would like to try to build on 10.10. When can you make the system  libjpeg and libpng default?

doutriaux1:
[...]

PR #1714, opened 2015-12-01
[WIP] PNG library is required on Mac

aashish24:
@dnadeau4 looks like on Mac we need PNG. I remember we had some discussions in the past on this issue. Can you remind me again? I tried finding it under the issue tracker but couldn't find anything.


aashish24:
@doutriaux1 Please have a look as well.

aashish24:
I should mention that this is WIP and not finalized yet.

jbeezley:
Need to be careful.  Using non-system libpng broke the build before because `xcode-select` is dynamically linked with the system version.  See https://github.com/UV-CDAT/uvcdat/issues/1012.

aashish24:
> Need to be careful. Using non-system libpng broke the build before because xcode-select is dynamically linked with the system version. See #1012.

thanks @jbeezley for reminding this. My understanding is that there is no system PNG library on Mac but folks install using brew.

jbeezley:
Actually, I think the problem is that it has the library in `/System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO` but not the headers.

jbeezley:
```
$ otool -L /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO
/System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO:
	/System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate (compatibility version 1.0.0, current version 4.0.0)
	/System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics (compatibility version 64.0.0, current version 600.0.0)
	/System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices (compatibility version 1.0.0, current version 48.0.0)
	/System/Library/PrivateFrameworks/AppleVPA.framework/Versions/A/AppleVPA (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1252.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1255.0.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 728.4.0)
	/System/Library/PrivateFrameworks/AppleJPEG.framework/Versions/A/AppleJPEG (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libJPEG.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libTIFF.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libPng.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libGIF.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libJP2.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libRadiance.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/System/Library/Frameworks/IOSurface.framework/Versions/A/IOSurface (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1225.1.1)
```

aashish24:
> Actually, I think the problem is that it has the library in /System/Library/Frameworks/ImageIO.framework/Versions/A/ImageIO but not the headers

that's worse. @chaosphere2112 @doutriaux1 how did you install libpng on your systems?

chaosphere2112:
Install XQuartz; it's listed in the [install instructions for mac](https://github.com/UV-CDAT/uvcdat/wiki/Install-on-Mac-OS-X)

dnadeau4:
We need PNG to read GRIB2 files with  g2clib.  Grib2 is using the PNG compression procedure to store data.   There is another PNG library from apple, but a wrapper will need to be written.  @doutriaux1 has been working on it in the pat.

aashish24:
> Install XQuartz; it's listed in the install instructions for mac

@chaosphere2112 yes, I installed it after realizing that it may have the libPNG but to me this solution is not quite right either as it seems like it will conflict with system PNG? unless it only install headers. Also, we don't need XQuartz for command line only options. I am looking for a more general solution that will work for command line and gui both.

doutriaux1:
I know this is a real pain! I had a branch where I was trying to include the updated png.h for macs. The issue is that of course mac changed all the call names so we need to add a prefix in headers... I will try to find the old branch but it might be dead when I `upgraded` my mac...

doutriaux1:
Why do we need that for? Xcode is a requirement now to fix this issue as far as i remember, anyhow the branch I was referring to is: https://github.com/UV-CDAT/uvcdat/tree/matplotlib_no_quartz_mac

doutriaux1:
@aashish24 look at my branch I had started the work of `#ifdef`ing the headers so that they work with system mac. Not sure how much was pushed to that branch though, I was close at one point and then gave up under he pressure of the `just install XQuartz` users.

doutriaux1:
@aashish24 it is reasonable to ask users to install XQuartz, most of them will ssh into a Linux machine anyway

aashish24:
> @aashish24 it is reasonable to ask users to install XQuartz, most of them will ssh into a Linux machine anyway

@doutriaux1 if I am not using x, why would I install XQuartz and corrupt my system? I don't think folks will like this. We shouldn't enforce libraries on users if they don't need it. If the issue is that we need libpng, then we should build it ourselves or install the headers only so that we can use system png. I am not concerned about 2.4 for 3.0, we should not depend on it.

dnadeau4:
@aashish24  Why should I install Xcode and corrupt my system.  If you want to finish Charles work to use libPng you are welcome to do so.  We work hard to add libpng this summer.  I am not changing it 3 months later.

aashish24:
> If you want to finish Charles work to use libPng you are welcome to do so.

Thanks, yes that's what I am thinking. 

> @aashish24 Why should I install Xcode and corrupt my system

@dnadeau4 didn't understand this sentence. Could you please elaborate? We need to install xcode and command line tools as they provide system packages that are first hand supported by Apple. Apple has deprecated use of X. Using X is problematic and we only need it for matplotlib plots (and hopefully even that will go away). Also, if XQuartz provides a libpng (header + library) even that is not ideal since a user could install libpng via brew. I am not against using XQuartz but I am not a big fan of using X

jbeezley:
Homebrew libraries that are duplicates of Mac system libraries are all marked [keg-only](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/FAQ.md#what-does-keg-only-mean) for this very reason.  They are never added to `/usr/local/lib` (unless the user forces it against recommendation) so you have to link against them explicitly.

aashish24:
> Homebrew libraries that are duplicates of Mac system libraries are all marked keg-only for this very reason. They are never added to /usr/local/lib (unless the user forces it against recommendation) so you have to link against them explicitly.

@jbeezley this is interesting. If they are duplicate (same version everything same), what would be the problem against using them for headers? or even linking. As per my understanding the system png does not provide headers as you pointed your in your comment.

jbeezley:
They are not the same version at all.  Mac OS has it's own non-libpng compatible API for reading images.  It just happens to include libpng internally for some old (probably modified internally somehow) version of libpng.  I think the reason that XQuartz works is that it installs libpng into a location that is not in the run time search path, so system binaries don't pick it up.  It also happens to install it into a location that the cmake/autoconf scripts find automatically.  When we install a custom libpng, it goes in LD_LIBRARY path and overrides the runtime search path for `xcode-select`.

The developers of homebrew make sure to never put third party libraries in front of system libraries in the search path ever.  This is (IMO) the primary benefit of homebrew over macports (or fink if that still exists).  Homebrew works with system OSX so you don't need to build an entirely new (pure GNU) toolchain from scratch.  I think the issue we are having is that we are trying to do the homebrew thing by using the OSX toolchain while putting our own libraries in front of system libraries.

aashish24:
@dnadeau4 @jbeezley

aashish24:
I am closing this for now since we force users to install png.

doutriaux1:
@aashish24  I'm restoring the branch there's quite some work in it if that's the one I remember!

doutriaux1:
ooops just looked at it, yes nothing important in this, redeleting the branch....

PR #1716, opened 2015-12-02
Made colormaps skip default indices

chaosphere2112:
@doutriaux1 Sorry, I lied. Found one more thing.

Matplotlib colormaps don't start with black like ours do; so let's skip 1-5.

```
import vcs, cdms2, os

cltfile = cdms2.open(os.path.join(vcs.sample_data, "clt.nc"))
clt = cltfile("clt")

canvas = vcs.init()

colormap = vcs.matplotlib2vcs("Blues")
canvas.setcolormap(colormap)
canvas.plot(clt)
canvas.png("matplotlib_colormaps")
```
results in 
matplotlib_colormaps

This should now be fixed.

### DIFF
--- a/Packages/vcs/Lib/colors.py
+++ b/Packages/vcs/Lib/colors.py
@@ -28,7 +28,7 @@ def matplotlib2vcs(cmap, vcs_name=None):
             (vcs_name, vcs_name_final))
     vcs_cmap = vcs.createcolormap(vcs_name_final)
     cmap_rgbs = cmap(range(0, cmap.N))
-    for i in range(5, 256):
-        vcs_cmap.setcolorcell(i, *([int(x * 100) for x in cmap_rgbs[(i - 5) % cmap.N][:4]]))
+    for i in range(5, min(cmap.N, 256)):
+        vcs_cmap.setcolorcell(i, *([int(x * 100) for x in cmap_rgbs[(i - 5)][:4]]))
 
     return vcs_cmap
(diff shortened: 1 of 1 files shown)


Packages/vcs/Lib/colors.py
```diff
@@ -28,7 +28,7 @@ def matplotlib2vcs(cmap, vcs_name=None):
             (vcs_name, vcs_name_final))
     vcs_cmap = vcs.createcolormap(vcs_name_final)
     cmap_rgbs = cmap(range(0, cmap.N))
-    for i in range(0, min(cmap.N, 256)):
-        vcs_cmap.setcolorcell(i, *([int(x * 100) for x in cmap_rgbs[i][:4]]))
+    for i in range(5, min(cmap.N, 256)):
+        vcs_cmap.setcolorcell(i, *([int(x * 100) for x in cmap_rgbs[(i - 5)][:4]]))
```
aashish24: setcolorcell first argument will be 5 now? Is that intended.

doutriaux1:
@chaosphere2112 no we talked about it with @williams13 1 through 16 and 240-255 used to be protected for this purpose, but in order to bypass this issue we implemented "string" colors i.e. `red` is an acceptable colors. What we might want to do is edit default templates to use `black`

chaosphere2112:
@doutriaux1 OK, I'll edit the default templates then.

Issue #1720, opened 2015-12-03
Data is clipped in interactive mode


danlipsa:
Further more, data is OK when saving a PNG but annotations are off.
Also, data is scaled differently in interactive mode compared with the PNG.

I used the following script provided by @aashish24 

#!/usr/bin/env python
import cdms2, cdutil, genutil
import vcs

if __name__ == '__main__':
    import sys
    cdmsfile = cdms2.open('~/build/uvcdat/install/share/uvcdat/sample_data/clt.nc')
    clt = cdmsfile('clt')
    clt = clt(latitude=(-90.0, 90.0),squeeze=1,longitude=(-180.0, 175.0),time=('1979-01', '1988-12'),)
    axesOperations = eval("{'latitude': 'def', 'longitude': 'def', 'time': 'def'}")
    for axis in list(axesOperations):
        if axesOperations[axis] == 'sum':
            clt = cdutil.averager(clt, axis='(%s)'%axis, weight='equal', action='sum')
        elif axesOperations[axis] == 'avg':
            clt = cdutil.averager(clt, axis='(%s)'%axis, weight='equal')
        elif axesOperations[axis] == 'wgt':
            clt = cdutil.averager(clt, axis='(%s)'%axis)
        elif axesOperations[axis] == 'gtm':
            clt = genutil.statistics.geometricmean(clt, axis='(%s)'%axis)
        elif axesOperations[axis] == 'std':
            clt = genutil.statistics.std(clt, axis='(%s)'%axis)
    canvas = vcs.init()
    gmIsofill = vcs.getisofill('a_robinson_isofill')
    args = []
    args.append(clt)
    gmIsofill.datawc_calendar = 135441
    gmIsofill.datawc_timeunits = 'days since 2000'
    gmIsofill.datawc_x1 = 1e+20
    gmIsofill.datawc_x2 = 1e+20
    gmIsofill.datawc_y1 = 1e+20
    gmIsofill.datawc_y2 = 1e+20
    gmIsofill.levels = [[1.0000000200408773e+20, 1.0000000200408773e+20]]
    gmIsofill.ext_1 = False
    gmIsofill.ext_2 = False
    gmIsofill.fillareacolors = [1]
    gmIsofill.fillareaindices = [1]
    gmIsofill.fillareastyle = 'solid'
    gmIsofill.legend = None
    gmIsofill.missing = 241
    gmIsofill.projection = 'robinson'
    gmIsofill.xaxisconvert = 'linear'
[...]

PR #1723, opened 2015-12-03
Clipped data

danlipsa:
Fixes for issue #1720

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -1223,17 +1223,17 @@ class VTKVCSBackend(object):
                 pt = vtk.vtkPoints()
                 Xrg2 = [1.e20, -1.e20]
                 Yrg2 = [1.e20, -1.e20]
-                if geo.GetDestinationProjection().GetName() in ["aeqd", ]:
+                if geo.GetDestinationProjection().GetName() in ["aeqd", "robin", "moll"]:
                     # These need more precision to compute actual range
                     Npts = 250
                 else:
                     Npts = 50
                 NGridCover = 0
                 pt.SetNumberOfPoints(Npts * Npts)
-                for x in numpy.arange(
-                        Xrg[0], Xrg[1], (Xrg[1] - Xrg[0]) / Npts):
-                    for y in numpy.arange(
-                            Yrg[0], Yrg[1], (Yrg[1] - Yrg[0]) / Npts):
+                for x in numpy.linspace(
+                        Xrg[0], Xrg[1], Npts, endpoint=True):
+                    for y in numpy.linspace(
+                            Yrg[0], Yrg[1], Npts, endpoint=True):
                         pt.InsertPoint(NGridCover, x, y, 0)
                         NGridCover += 1
                 pts = vtk.vtkPoints()
--- a/testing/checkimage.py
+++ b/testing/checkimage.py
@@ -51,7 +51,7 @@ def find_alternates(fname):
     return results
 
 def check_result_image(fname, baselinefname, threshold = defaultThreshold,
-                       baseline = False, cleanup=True):
+                       baseline = True, cleanup=True):
     testImage = image_from_file(fname)
     if testImage is None:
         print "Testing image missing, test failed."


aashish24:
Nice @danlipsa!

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -1231,9 +1231,9 @@ def fitToViewport(self, Actor, vp, wc=None, geo=None, priority=None,
                 NGridCover = 0
                 pt.SetNumberOfPoints(Npts * Npts)
                 for x in numpy.arange(
-                        Xrg[0], Xrg[1], (Xrg[1] - Xrg[0]) / Npts):
+                        Xrg[0], Xrg[1] + (Xrg[1] - Xrg[0]) / Npts, (Xrg[1] - Xrg[0]) / Npts):
```
aashish24: @danlipsa can you explain what's going on here?

aashish24:
@danlipsa you would have to update all the baselines as well I believe. The workflow is that you upload your baselines first and then push the code-branch. since then the buildbots can pick up the right data branch corresponding to the code-branch. Hope this helps.

Packages/vcs/Lib/VTKPlots.py
danlipsa: Here we generate a 2D grid of points to cover the whole domain and transform those points through the geographic projection. The goal is to get the bounds of the projection. The problem is that arrange uses an interval open at the higher end, so the points on the margin are not included. What I have done is to include the points on the margin.

Of course a better fix would be to pass the bounds of the projection but that will take more time to learn the code.

aashish24:
thanks so if Xarg[1] say is 100, Xarg[0] = 60 and N=50

Then Step = 40/50 = 0? in numpy.arrange? and also 

the arguments to numpy.arrange would be:
(60, 100, 0) Is that correct? which would be open-ended [65, 100)

- Aashish

danlipsa:
I think those are floats so the code uses the correct step.

aashish24:
> I think those are floats so the code uses the correct step.

Ah.. I didn't see that. Thanks, yes, then your change make sense to me and looks correct. Are you going to update the baselines as well?

aashish24:
Ref: https://github.com/UV-CDAT/uvcdat/issues/1725

danlipsa:
@aashish24 yes I wll. I will also add a test for the problem I am solving. I had some issues with a couple of tests producing empty values, so in that case I reverted to what we were doing before. Strangely that was aeqd actually working correctly. I have to see what was the difference between that test and the test that does not work.

aashish24:
@danlipsa the test images are looking much better and correct and that's why they are not matching up with baselines. I would suggest you update the baselines as well. I don't know why garant is not showing these test failures.

doutriaux1:
@danlipsa see: https://open.cdash.org/viewTest.php?onlyfailed&buildid=4138094 newly generated pix seem better than baseline, I would recommend updating.

danlipsa:
@doutriaux1 We found two more projections that needed more points with the new changes. Those were robinson and  mollweide only for meshfill. Do you remember why we compute the bounds of the geographic projection instead of passing it as a parameter?

danlipsa:
@doutriaux1 @aashish24 I would get rid of that computation altogether but probably it is too involved to do it before the release.

doutriaux1:
@danlipsa you're right we should pass it as a parameter, `projection_type:array` in case the projection type changes between refresh. Let's take a look post 2.4

aashish24:
@danlipsa did you update all the baselines as we discussed today?

aashish24:
@williams13 @doutriaux1 we found few issues in the code, hoping that this branch will fix some of them. For others we need to discuss.

danlipsa:
@aashish24 @doutriaux1 I did. It seems that at least for the first file this is a difference between PNG generated on my machine and PNGs generated with MESA. It seems the difference is the black line at the top. Do you guys know what that is?

danlipsa:
@aashish24 @doutriaux1 Note the Mesa generated file has the black line at the bottom. Should we increase the threshold?

doutriaux1:
it's a bit disturbing, the black line is in a layer higher than the data, it should be on top of it. VTK bug? MESA bug?

aashish24:
@danlipsa if you are talking about test_polar_set_opt_param_polar.png then I would say that in the old baseline the continents are clipped on top bottom left and right. So the new baseline seems correct to me.

danlipsa:
@doutriaux1 @aashish24 I think it is fairly irrelevant for the correctness of the results shown if the black line is there or not. I would suggest adding another baseline for these tests. My guess is that there is a difference in the OpenGL drivers for Mesa. I would advise adding another baseline. Note that the tests on   cont-int/LLNL/Darwin-Mac (NOGUI) have all passed. We can also add an issue for someone to look into this after the release, but I don't think we should delay the release for this. Thoughts?

danlipsa:
@aashish24 We are talking about the test results on Mesa
https://open.cdash.org/viewTest.php?onlyfailed&buildid=4138409

aashish24:
thanks @danlipsa  @doutriaux1 are we drawing these boundaries? 

@danlipsa these lines were in the old baseline. If you upload the new baselines, your tests should pass?

danlipsa:
@aashish24 I think the lines were not in the old baselines. I think we do draw them. Maybe the layer with the lines is close enough to the data that Mesa draws is under the data. Note that both my machine and the tests in cont-int/LLNL/Darwin-Mac (NOGUI) shows the lines (these have hardware OpenGL). We did change slightly the parallel projection and probably that made the difference.

danlipsa:
@doutriaux1 Are you saying that the black line should be there? This is what I got with OpenGL on my machine.

doutriaux1:
@aashish24 @danlipsa fyi the competition using proj4 better than us.... http://basemaptutorial.readthedocs.org/en/latest/projections.html

aashish24:
@danlipsa changes looks good to me. 

@doutriaux1 yes, we are going to use latest proj4 and we can use our own. I have a small fix for VTK for that. The only thing we need is the mangled proj4 for VTK.

aashish24:
@doutriaux1 can we merge this branch now?

doutriaux1:
@aashish24 let me run the test suite one last time before merging.

aashish24:
sure @doutriaux1 but what are you concerned about? all tests are passing on garant and mac and the code looks reasonable.

doutriaux1:
just want to add the extra step. There's a lot of yellow on the dashboard.

aashish24:
@doutriaux1 are you talking about buildbots not running?

doutriaux1:
yes @aashish24 mac are slowly catching up, RH need to restart (I really need to figure out which test makes them hang!)

aashish24:
@doutriaux1 any luck?

doutriaux1:
still need to run ctest connection keeps dropping here....

aashish24:
Are you running ctest locally? or are you trying to start the Linux buildbots?

doutriaux1:
locally on my machine at the lab, which means remote X forwarding which is flaky here.

aashish24:
Okay... are you concerened that dashboards are not catching something (garant and mac)..

doutriaux1:
just want to be sure nothing showed up on Lniux, will probably try to retrigger the bots as well.

aashish24:
Oh okay. .. garant seems to be happy. Let me know and I can run tests on my machine as well (in case you are having issues).

doutriaux1:
@aashish24 it was good to wait on the bots. @danlipsa can you please up a bit the thresholds

danlipsa:
@doutriaux1 Do you have the files added by this commit in your mesa buildbot?
https://github.com/UV-CDAT/uvcdat-testdata/commit/d1ef16da34b68a58641ec327b3bc62d5a225a40e
It seems that the errors we get are the same as before I added these files.

aashish24:
@doutriaux1 looks like your bot are not picking the _1 files. Can you check if these files are downloaded on the failing bots?

aashish24:
@danlipsa I updated the branch with current master to see if LLNL bots will pick up update _1 images. If not we would have to work on the threshold for now as @doutriaux1 suggested.

doutriaux1:
@danlipsa do you still need me to merge this or should I wait for #1736 ?

doutriaux1:
@danlipsa it's all good on my ubuntu here.

danlipsa:
@doutriaux1 It would be great if you could merge this. I can rebase #1736 after that.

doutriaux1:
ok so what about the bots? Why are they failing can we up the thresholds?

aashish24:
> ok so what about the bots? Why are they failing can we up the thresholds?

@doutriaux1 I guess you didn't read all the messages. On MESA we are seeing this extra line. I believe it could be because of the polygon offset not working or set on MESA. To resovle this issue for now @danlipsa uploaded _1 versions but seems to me that they are not getting picked up by bots at LLNL. Can you chekc if the data checkout has _1 images? @danlipsa did you try what we discussed this morning? Other thing @danlipsa we could do is set the global offset in VTK to have the lines always show up on top?

doutriaux1:
@aashish24 I did read the thread and I did reply the line `should` always be on top! It's the same priority as the data and drawn later. So that needs to be addressed. I will check to see if the _1 are picked on bots.

doutriaux1:
@aashish24 @danlipsa we can solve that line issue post 2.4. But we need the bots to be reasonably ok before the release. Will check on the _1 on bots.

aashish24:
oh okay thanks @doutriaux1. @danlipsa try setting the polygonoffset? (or lineoffset?)

danlipsa:
This has new baselines
https://github.com/UV-CDAT/uvcdat-testdata/pull/92

danlipsa:
@doutriaux1 @aashish24 I tried to switch the regular png with the _1 png for one of the tests to make sure ctest works correctly. It does. The test still passed on my machine even if the png that matched was the _1 one.

danlipsa:
@doutriaux1 @aashish24 So, it does appear that maybe the buildbots do not pull the _1 files correctly.

doutriaux1:
@danlipsa I just double check after a retrigger the _1 files are here for oceanonly ( http://aims4.llnl.gov:9981/master/macbot2.llnl.gov:9982/UV-CDAT/uvcdat/8055818c0a23e7019b638dbb2ed7e49eb95a7999/ctest__-j4__-D )

I'm not sure what you're trying to do ,but the files are defintely there, what are the _1 for?

The way the process work on the bot is that master is checked out then the branch `clipped-data` then master is merged back in, would that throw off your test?

doutriaux1:
runnnig the ctest on your branch with uvcdat-data having been subjected to a:  `git reset --hard origin/clipped-data`. Will post results soon.

doutriaux1:
@aashish24 @danlipsa still fails all the test, will post the link when done. What are the _1 for?

doutriaux1:
Could it be due changes in master that are put back into your branch? Can you trst on your computer after you merged master back into `clipped_data`?

doutriaux1:
okay failing link: https://open.cdash.org/viewTest.php?onlyfailed&buildid=4148428

danlipsa:
@doutriaux1 The _1 files are alternate baselines. ctest looks for alternate baselines if the main image fails the test.

danlipsa:
@doutriaux1 @aashish24 I could recompile VTK with Mesa on my machine and try that. Hopefully it will have the same behavior as the test machine.

aashish24:
@danlipsa why don't we set line offset so that lines are always drawn on top. May be it will take care of some of the issues. I don't think it will take care of all issues..

danlipsa:
@doutriaux1 Can you try running the first test that fails from command line:vcs_test_basic_meshfill_aeqd_proj_SH_gm Does it fail?
If yes, can you try copying the _1 baseline over the regular baseline:
cp test_vcs_basic_meshfill_aeqd_proj_SH_via_gm_1.png test_vcs_basic_meshfill_aeqd_proj_SH_via_gm.png

Run the test again. Does it fail?
This will give us more information about what is wrong with the test machine or with the test.

danlipsa:
@aashish24 I will try looking at this maybe I see something. The problem is that I don't have the machine where the line is not drawn so I have no way to test.

danlipsa:
@aashish24 @doutriaux1 I'll use OSMesa maybe this will have the same behavior as the test machine.

danlipsa:
@aashish24 @doutriaux1 I compiled with OSMesa and indeed it does not seem to pickup the _1 file. I will look into why.

doutriaux1:
@danlipsa @aashish24 I'm still perplexed by this `_1` business. I'd say increase the threshold for the release and let's try to fix it properly post-release. Thoughts?Although MESA seems to be the only one that does things correctly with the line on top. So it may be a VTK or opengl issue.

danlipsa:
@doutriaux1 @aashish24 The reason why the test would not pickup the _1 is a parameter to checkimage.check_result_image that said it should not. See my latest commit. Hopefully this will make all tests pass. I can keep looking into why the line does not show up on OSMesa but if all the tests pass we can release before that.

danlipsa:
@doutriaux1 @aashish24 Ready to merge? 5/9 are green and it has been running for a while. Not sure if anything else is going to finish.

danlipsa:
@aashish24 @doutriaux1 Looking at why on OSMesa we don't have line showing, it seems that this is a fitToViewport issue as well rather then a polygonoffset/lineoffset issue. Given that the parallel projection depends on the matrix of dots we generate it makes sense that this problem showed up when I changed the matrix of points in this bug fix. My opinion is that we should look into passing the actual bounds of the projected dataset. After that we will see if we still have the line problem on OSMesa. Thoughts?

aashish24:
> as well rather then a polygonoffset/lineoffset issue. Given that the parallel projection depends on the matrix of dots we generate it makes sense that this problem showed up when I changed the matrix of points in this bug fix. My opinion is that we should look into passing the actual bounds of the projected dataset. After that we will see if we still have the line problem on OSMesa. Thoughts?

+1 can you file a bug for this @danlipsa 

@danlipsa the dashboards are looking pretty good now.. failures on garant seems to be an issue related to not pulling in the right test data branch. May be you need to update your data branch as well. But other than that.. its +1 for me. @doutriaux1 ?

danlipsa:
@aashish24 See https://github.com/UV-CDAT/uvcdat/issues/1739

danlipsa:
@doutriaux1 Is this ready to merge?

aashish24:
dan lets merge this branch. Any remainig concerns if not found or discussed can be handled in another pr.

Issue #1354, opened 2015-05-28
Update license/legal info


durack1:
The current [legal/license info](https://github.com/UV-CDAT/uvcdat/blob/master/docs/Legal.txt) was generated when LLNL was managed by UC.

This likely needs updating, and streamlining to a single point of reference for all license related information. We currently have [Contributor license](http://uvcdat.llnl.gov/governance.html) documents on the web (bottom of the page), but there doesn't appear to be an equivalent "Usage license" available.

@williams13 - if required it may be a good time to revisit this with the current LLNS legal team. If this occurs, it would be useful to know whether the software needs to be distributed with license info (within the tar ball) or whether a URL pointing to the relevant info in `README.md` is sufficient

durack1:
@doutriaux1 @mattben I get the impression this issue isn't going to be resolved in the `2.4.0` timeframe - I suggest shifting this to the `3.0` milestone..?

aashish24:
@durack1 done.

durack1:
@aashish24 I'm not sure the legal update is done - take a look at the first link above..

aashish24:
@durack1 thanks!

mattben:
@doutriaux1 What do you need from me on this issues (which is quite old) is there a new License for CDAT v8 to put on the website? Which comes from the CDAT Repo

PR #1727, opened 2015-12-08
Moved/renamed license

aashish24:


aashish24:
@williams13 @charlesdoutriaux should we use the Apache V2 license instead of using the one we have?

aashish24:
@danlipsa please review

doutriaux1:
@aashish24 not sure we can, I think the lab requires this one. What is the advantage of Apache2?

chaosphere2112:
http://www.apache.org/licenses/

williams13:
Let’s discuss this next week.

-Dean

LICENSE
```diff
@@ -1,5 +1,5 @@
-CCLRC Licence for CCLRC Software forming part of the 
- Climate Data Analysis Tools Package
+CCLRC License for CCLRC Software forming part of the  Ultrascale Visualization and
```
doutriaux1: I don't think there's a `and` here @williams13 ?

aashish24:
@doutriaux1 please have a look here: http://choosealicense.com/

The apache license is nice since it is well defined and just a lit bit restrictive than bsd for good reasons. Also, it is compatible with GPL3 (but not GPL2). It takes care of patents issues (not sure if this will affect us). NASA encourages Apache.

durack1:
@aashish24 @doutriaux1 @chaosphere2112 @williams13 I'd note there is an open issue about this #1354 - would be best to link these so as the issue is resolved it's closed.

aashish24:
Folks  I am merging this one in once we discuss the issue more in detail at AGU or past AGU, we can update the branch. Thanks to @durack1 for referring to the existing issue.

PR #1731, opened 2015-12-10
Issue 1730 mesh with missing vertices

doutriaux1:
@aashish24 @sankhesh thanks to @sankhesh help I got the shapes to be drawn correctly, but they still appear to be drawn as missing, I think the "nan" bit throws off VTK ? Help wanted :wink:

### DIFF
--- a/Packages/vcs/Lib/vcs2vtk.py
+++ b/Packages/vcs/Lib/vcs2vtk.py
@@ -89,7 +89,6 @@ def putMaskOnVTKGrid(data, grid, actorColor=None, cellData=True, deep=True):
             lut = vtk.vtkLookupTable()
             r, g, b, a = actorColor
             lut.SetNumberOfTableValues(2)
-            print "CELL DATA:",cellData
             if not cellData:
                 grid2.GetPointData().RemoveArray(
                     vtk.vtkDataSetAttributes.GhostArrayName())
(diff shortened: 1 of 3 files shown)


aashish24:
great @doutriaux1 how should we handle nan?

doutriaux1:
@aashish24 basically nan means `ignore this vertice` that's how we handle grids which have some cells with a different number of vertices between them (3 vs 4 here) we use the greatest number of vertices and set the unused ones to missing, which @sankhesh suggested we convert to `nan` for VTK. Unfortunately VTK does pick the correct shape now but seems to mark the cell has `bad`

doutriaux1:
@aashish24 @sankhesh @chaosphere2112 ready for review

doutriaux1:
goes with: https://github.com/UV-CDAT/uvcdat-testdata/pull/94

aashish24:
@doutriaux1 if this is a uniform grid, we should probably using masking array. Thoughts?

chaosphere2112:
@aashish24 Actually, this is for `meshfill`, which is inherently a non-uniform grid.

doutriaux1:
@aashish24 it is actually using masked arrays anyway. And @chaosphere2112 is right these are non-uniform grids, and in this case really non-uniform (different number of vertices for each cell)

aashish24:
@doutriaux1 thanks.. can you fix the style issue?

doutriaux1:
nah...

aashish24:
@chaosphere2112 sorryI meant to say regular grid / structured grid which curve linear is.

Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -263,7 +263,7 @@ def genGrid(data1, data2, gm, deep=True, grid=None, geo=None):
                 xM = m[:, 1].max()
                 ym = m[:, 0].min()
                 yM = m[:, 0].max()
-                N = m.shape[0]
+                numberOfCells = m.shape[0]
```
aashish24: nice var name thanks :+1:

aashish24:
@doutriaux1 changes LGTM :+1:

aashish24:
If bots are happy, I am going to merge it.. nice job @doutriaux1 @sankhesh

doutriaux1:
@aashish24 please do not forget to merge baselines as well. Thanks. I just merged it in. Thx @durack1 for spotting this.

PR #1735, opened 2015-12-11
Issue 1734 non contiguous levs

doutriaux1:
@sankhesh @aashish24 @chaosphere2112 please review

### DIFF
--- a/Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
+++ b/Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
@@ -5,7 +5,6 @@ import fillareautils
 import numpy
 import vcs
 import vtk
-import isofillpipeline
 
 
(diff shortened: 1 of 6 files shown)


doutriaux1:
goes with: https://github.com/UV-CDAT/uvcdat-testdata/pull/95

aashish24:
@doutriaux1 this will some time to review.. more than minor change of code.. 

@danlipsa it would be nice if you can look into this branch as well.

doutriaux1:
@aashish24 it is really minor, I just unified the function call, it was the same function called from isofill/boxfill/meshfill so I moved it to the mother class instead.

aashish24:
ah..thanks @doutriaux1 looking into it..

aashish24:
@doutriaux1 looks good to me.. crunchy is having some serious issues.. can you have a look at it? 

Thanks

PR #1752, opened 2016-01-04
several tweaks, mainly saving method results for re-use, to improve

doutriaux1:
performance of climatology etc. on Rhea or other machines where I/O
is expensive.

WORK IN PROGRESS DO NOT MERGE YET


Packages/cdms2/Lib/axis.py
```diff
@@ -735,10 +735,22 @@ def designateTime(self, persistent=0, calendar=None):
             if calendar is not None:
                 self.setCalendar(calendar, persistent)
 
+    # For isTime(), keep track of whether each id is for a time axis or not, for better performance.
+    # This dictionary is a class variable (not a member of any particular instance).
+    idtaxis = {}  # id:type where type is 'T' for time, 'O' for other
+
```
doutriaux1: @painter1 what happens if you have mutliple files with same axis id, but only some are actually time? Wouldn't this break?

painter1: Yes, it would break!  I've never seen such a thing, but among the changes I've made, this is the one which I was least sure about.

doutriaux1: maybe we should add a test? See if it's already in and if it is and type changes, throw a warning/error? I don't think it will ever happen but of course now that I said it it is bound to happen...

painter1: The point of this is to not have to figure out what the type is - that involves reading from a file.  So the way I would fix it is to use self.idtaxis only if the user does something to enable this feature, e.g. calls a method to initialize it.

aashish24:
I will have a look at it for review.

painter1:
Is this good enough to be merged?   We need this for release 2.4.1, and the code freeze is tomorrow.

aashish24:
@painter1 it looks good to me, certain code I couldn't reasonably test since its very specific but overall its good. 

+2

aashish24:
Our Mac bots are hosed but I am okay if we merge it as it is.

PR #1753, opened 2016-01-04
Axis labels now move when scaled by ratios

chaosphere2112:
Added `xname` and `yname` to the series of template attributes modified by setting the ratio. They are moved to a scaled-down percentage of the original difference between themselves and the corresponding `(x|y)label1` attribute.

The script:

```python
import vcs, cdms2
f = cdms2.open(vcs.sample_data + "/clt.nc")
s = f('clt')
x = vcs.init()
x.plot(s, ratio="autot")
```

Here's the original:

bad

If you shrank the width of the window, a similar thing would happen with the xname:

tall_bad

After this PR:

nice

and (still a bad plot, but the label is in the right place):

tall_good

### DIFF
--- a/Packages/vcs/Lib/Canvas.py
+++ b/Packages/vcs/Lib/Canvas.py
@@ -866,9 +866,7 @@ class Canvas(object):
         vcs.next_canvas_id += 1
         self.colormap = None
         self.backgroundcolor = 255, 255, 255
-        # default size for bg
-        self.bgX = 814
-        self.bgY = 606
+
         # displays plotted
         self.display_names = []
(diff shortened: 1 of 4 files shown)


chaosphere2112:
Fixes #1624

doutriaux1:
@chaosphere2112 should we add a test for `ratio='autot'` ?

chaosphere2112:
@doutriaux1 Yeah, probably. I'll do one using the examples above.

chaosphere2112:
Test data: UV-CDAT/uvcdat-testdata#99

PR #1772, opened 2016-01-07
BUG: proj4 over option causes problems with polar projections

danlipsa:
On certain machines, +over option to proj4 results in wrong
projections. It is not clear why there is a difference between different
computers - we have seen this behavior on both unix and mac. +over
does not wrap points outside of -180, 180. This is not necessary for
polar projections so we remove this option in that case.

### DIFF
--- a/Packages/vcs/Lib/vcs2vtk.py
+++ b/Packages/vcs/Lib/vcs2vtk.py
@@ -570,24 +570,22 @@ def projectArray(w, projection, wc, geo=None):
         pname = projDict.get(projection._type, projection.type)
         projName = pname
         pd.SetName(projName)
-        if projection.type == 'aeqd':
-            # this is a temporary branch to keep the same
-            # baselines
+        if (projection.type == 'aeqd' or
+                projection.type == 'polar stereographic'):
             setProjectionParameters(pd, projection)
+        elif projection.type == "polar (non gctp)":
+            if ym < yM:
+                pd.SetOptionalParameter("lat_0", "-90.")
+                pd.SetCentralMeridian(xm)
+            else:
+                pd.SetOptionalParameter("lat_0", "90.")
+                pd.SetCentralMeridian(xm + 180.)
         else:
             pd.SetOptionalParameter("over", "true")
-            if projection.type == "polar (non gctp)":
-                if ym < yM:
-                    pd.SetOptionalParameter("lat_0", "-90.")
-                    pd.SetCentralMeridian(xm)
-                else:
-                    pd.SetOptionalParameter("lat_0", "90.")
-                    pd.SetCentralMeridian(xm + 180.)
-            else:
-                setProjectionParameters(pd, projection)
-                if ((not hasattr(projection, 'centralmeridian') or
-                     numpy.allclose(projection.centralmeridian, 1e+20))):
-                    pd.SetCentralMeridian(float(xm + xM) / 2.0)
+            setProjectionParameters(pd, projection)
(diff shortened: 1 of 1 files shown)


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/100

doutriaux1:
@danlipsa THANKS! I will do a bit of testing for lambert as well, maybe it's better to go with a simlar approach than the one we have for labels, create  a list of known projection that need `over` and only set it for these, that should limit the bug a bit further, no? see: https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/projection.py#L21

danlipsa:
Now I know some projections which don't work well with over: polar (non gctp), aeqd and polar stereographic. For polar projections you don't need over because when you wrap you end up in the same place - so that is no problem.

I think passing over to all projections should work - and it does on some machines :-)
So maybe we think about rearranging the code after we upgrade to the latest proj4 which may work more consistently.

doutriaux1:
ok, but lambert fails with over as well for sure.

doutriaux1:
that's a lot of elif, also with the restrcturing happening in lambert branch, it seem the gctp will not get the parameters autosets?

doutriaux1:
really why not simply check against a list of proj that do not need this parameters, seems like a little more flexible, let me try to implement it in my lambert branch and lets compare ok?

danlipsa:
Sure, no problem.

danlipsa:
Does lambert (on your branch) fails only on annie or it fails everywhere?

doutriaux1:
on annie and macs were the -3 tests fail. Wors when -3 works.

danlipsa:
@doutriaux1 @aashish24 I wonder if this is a compiler issue. This is one thing that is different between annie/my laptop and my desktop. Annie has gcc 5.2.1, my laptop has llvm 7.0.2 my desktop has gcc 4.8.4. The proj4 we are using is over 10 years old so the new version should have many fixes.

danlipsa:
@doutriaux1 @aashish24 In my mind +over should be always passed as proj4 wraps around only points which create problems with the cells of the dataset. So we should revisit this after we upgrade proj4 and try to always pass +over. (for polar projection an wrapped point ends up in the same place so it should not be a difference there)

aashish24:
@doutriaux1 can we merge this branch since its passsing on  anie and other machines and then fix the lambert?

aashish24:
@doutriaux1 I am merging this as I think it looks good to me.

doutriaux1:
no non no no no NOOOOOO! It's the wrong approach I talked about it with @danlipsa this does not solve anything.... I will revert this. Beside it doesn't fix anything the issue is in proj4, it still fails on macs....

danlipsa:
@doutriaux1 Which tests are failing? I have not seen anything in the buildbots.

doutriaux1:
when i implement this in my lambert branch, the lambert test gives a different plot every time I run it. 

Anyway the lambert+proj4_updated is almost done building on my mac which exposes all the issue, I have hope...

danlipsa:
@doutriaux1 @aashish24 I don't have anything against restructuring this code. If this code does not work with the lambert branch should be fixed/adjusted by that branch. I don't think that we should delay fixes for the master branch because those fixes don't fit some work in progress. If a fix passes all tests it should be good to go. Now lambert branch has to fix everything that this branch fixed plus what it was fixing. Plus we delay valid fixes for master.
In this particular case, as long as lambert goes in quickly and fixes everything, I don't have a problem with this. :-)

PR #1773, opened 2016-01-08
[WIP] Build and use newer version of proj4

aashish24:
Because of recent reports of issues with much older proj4, I am attempting to update to newer PROJ4 (Updated UV-CDAT/VTK as well).


aashish24:
Ref VTK: https://github.com/UV-CDAT/VTK/pull/14

aashish24:
@williams13 @doutriaux1 @danlipsa to overcome weird system specific issues with projection, I have pushed changed to VTK and UV-CDAT to use newer PROJ4.

aashish24:
@doutriaux1 @danlipsa I believe it didd help with the -3 but aeqd is a problem. May be we need to pass the correct parameters for aeqd.

danlipsa:
@aashish24 Note that -3 used to work on most systems anyway. Unless you have one of those systems it did not work on (annie, maybe newer macs - my laptop is one of them) we cannot definitely say it helps.

aashish24:
what is -3 projection (in terms of PROJ4 terms)? For the newer RPOJ4 it looks like they broke AEQD.

danlipsa:
@aashish24 -3 is the same as aeqd but with 90/-90 lat_0

aashish24:
> @aashish24 -3 is the same as aeqd but with 90/-90 lat_0

I see. So it seems like newer PROJ4 is bad for lat_0 lon_0 = 0

Do you want to look into it?

danlipsa:
Remember aeqd was broken to start with. So I would not say they broke aeqd but just that it is different then the old one. Sure, I can look into it. Are we still going to release with the old proj4? Do I finish wrapping of meshfill data first?

doutriaux1:
@aashish24 did you copy the dep files from my [ocgis](https://github.com/UV-CDAT/uvcdat/tree/ocgis) branch?
If not why update gdal? We don't use it except for ocgis?

aashish24:
> @aashish24 did you copy the dep files from my ocgis branch?
> If not why update gdal? We don't use it except for ocgis?

@doutriaux1 I copied some things from your branch. We can take out the gdal.

doutriaux1:
ok that is what it looked like, I'd say we should have just merged that branch in, but it's so close from the deadline I see why you didnt. I will take the gdal bits out before we erge this into master if that indeed fixes our issues.

doutriaux1:
proj4 build failed on annie (ubuntu/mesa)

aashish24:
> proj4 build failed on annie (ubuntu/mesa)

Right. I don't want to change anything too close (post the expected date of release). This was my attempt and thats why WIP to see what happens on Annie and other machines.

aashish24:
@doutriaux1 we can look into the OCGIS branch post 2.4.0 and we can just take my VTK branch in after we confirm that newer PROJ4 didn't build anything.

doutriaux1:
yes ocgis is 2.6

PR #1790, opened 2016-01-14
Add back the lambert test on macs.

danlipsa:


danlipsa:
@doutriaux1 @aashish24 It seems that the mac buildbots are hanged. Could someone restart them? Thanks!

doutriaux1:
I did I restart the lambert one but then killed rthem... my bad, restarting again.

danlipsa:
@doutriaux1 @aashish24 No tests were run on macs yet, everything else finished. Are macs slower?

doutriaux1:
@danlipsa fails on mac as expected see: https://open.cdash.org/testDetails.php?test=410161895&build=4190915

aashish24:
> @danlipsa fails on mac as expected see: https://open.cdash.org/testDetails.php?test=410161895&build=4190915

Looks like continents are off?

doutriaux1:
@danlipsa I added an extra step at the end the FULL did run but doesn't show it as ran... Taking that out right now. Look at the link I posted above.

doutriaux1:
@aashish24 it's the same proj4/vtk bug, everything is drawn completely off, everytime I run this test I get a different output. So it really looks like memory not being initialized properly or something is reading the wrong type in memory,too much of it, not enough of it, etc...

danlipsa:
@doutriaux1 @aashish24 Interesting. Looks like a fitToViewport coupled with an uninitialized memory issue. Note the contour/continents compute the bounds of the dataset using the array of points, the data uses the actual bounds from the dataset. I am working to replace the old fitToViewport (the one that does the computation) for continents and contour. I would be worth it to try again after I am done.

aashish24:
> @doutriaux1 @aashish24 Interesting. Looks like a fitToViewport coupled with an uninitialized memory issue. Note the contour/continents compute the bounds of the dataset using the array of points, the data uses the actual bounds from the dataset. I am working to replace the old fitToViewport (the one that does the computation) for continents and contour. I would be worth it to try again after I am done.

+1 

> @aashish24 it's the same proj4/vtk bug, everything is drawn completely off, everytime I run this test I get a different output. So it really looks like memory not being initialized properly or something is reading the wrong type in memory,too much of it, not enough of it, etc...

Okay, thanks for the update, I guess we can valgrand on it check on uninitialized values. I will do it once I am done with my sprint tasks.

danlipsa:
@aashish24 Sounds good. Let us know what you see. Wanted to to this as well just happened that I started working on fitToViewport so now I'll rather finish this before I do something else. I'll let you have fun with it. :-)

danlipsa:
image

Issue #1792, opened 2016-01-14
vcs_test_init_open_sizing fails on small monitors


chaosphere2112:
When I run  vcs_test_init_open_sizing  on a small laptop screen, the test fails. I believe this is the same issue that @williams13 ran into when testing the new build. I'll see what I can dig up.

aashish24:
thanks @chaosphere2112 I added it to the ready queue.

durack1:
@aashish24 is the `0 - backlog` through `4 - Done` the way you plan to better label issues/PRs? Is this documented somewhere?

chaosphere2112:
@durack1 They're a side effect of https://huboard.com/UV-CDAT/uvcdat , which we're trying out for project management.

PR #1793, opened 2016-01-14
Got rid of screen size logic from test

chaosphere2112:
Not entirely certain why I put it there in the first place; test should run perfectly fine with a window that's bigger than the monitor.

### DIFF
--- a/testing/vcs/test_vcs_init_open_sizing.py
+++ b/testing/vcs/test_vcs_init_open_sizing.py
@@ -46,12 +46,6 @@ def test_canvas_size(c, size, via):
     info = c.canvasinfo()
     w, h = size
 
-    # Make sure size fits on screen bounds
-    screen_w, screen_h = c.backend.renWin.GetScreenSize()
-    if w > screen_w:
-        w = screen_w
-    if h > screen_h:
-        h = screen_h
     assert info["width"] == w, "Width via %s incorrect; expected %d, got %d" % (via, w, info["width"])
     assert info["height"] == h, "Height via %s incorrect; expected %d, got %d" % (via, h, info["height"])
 


chaosphere2112:
Fixes #1792

durack1:
@chaosphere2112 probably had something to do with `OS X 10.8.old` or was it `Windows 3.1`:wink:

chaosphere2112:
@durack1

doutriaux1:
@chaosphere2112 the bots use to pass before anyway. Does it pass again on your ubuntu?

testing/vcs/test_vcs_init_open_sizing.py
```diff
@@ -46,12 +46,6 @@ def test_canvas_size(c, size, via):
     info = c.canvasinfo()
     w, h = size
 
-    # Make sure size fits on screen bounds
-    screen_w, screen_h = c.backend.renWin.GetScreenSize()
```
doutriaux1: I think GetScreenSize either works or not in VTK... Never could figure out when it does or not, so it probably a combination of this + small screen

durack1:
@chaosphere2112 @doutriaux1 just as a comment, if you are depending on a `vnc` configuration to test things, then I'd be pretty skeptical that it's representative of a physical machine configuration.. `vnc` provides super quirky results with graphics, trying to plot things etc etc.. Particularly if it's in a background mode..

testing/vcs/test_vcs_init_open_sizing.py
aashish24: @doutriaux1 if you remember @dlonie pushed a fix for this and now it should work on all systems.

aashish24:
@chaosphere2112 changes looks good to me. Yes, I agree that the test was not that useful since the initial screen size could be some number depending on the X or the underlying window system.

aashish24:
@doutriaux1 I am merging this.

PR #1794, opened 2016-01-15
Fixed bug that was breaking continents when set at canvas level

chaosphere2112:
Found a bug that wasn't properly covered by my test_continents script; if you set the continents at the canvas level, rather than by using the continents argument of the plot function, the continents would be ignored.


chaosphere2112:
@doutriaux1 Flake8'd

aashish24:
@chaosphere2112 changes looks good to me. LGTM :+1:

sankhesh:
LGTM :+1:

Issue #1795, opened 2016-01-15
ratio="autot" fails to correctly generate the initial ratio


chaosphere2112:
If you plot something with `ratio="autot"` as the very first action you take with a canvas, the plot defaults to using the default window size, rather than the size that the canvas will actually have. This is because the call to determine the ratio happens *prior* to any calls to `createRenWin()` in the backend, so default values are retrieved using the `canvasinfo()` function.

Any future plots (or ones triggered by `canvas.update()`) will be the correct proportions.

chaosphere2112:
I've sort-of got a fix for this worked out; it needs work, though. I'm running a full test suite on it and sending it out the the dashboard.

chaosphere2112:
@doutriaux1 https://open.cdash.org/viewTest.php?onlyfailed&buildid=4197349 Getting closer...

doutriaux1:
looks like the diags doesn't set the bg dims and throws the tests off.

chaosphere2112:
Does diags use Multi? Because almost all of the multi tests failed as well

chaosphere2112:
@doutriaux1 So Multi was using "x.portrait" and "x.landscape", which triggered a call to canvasinfo prior to plot, which meant that bg wasn't set. I'll wager it's something similar in the diagnostics.

https://open.cdash.org/viewTest.php?onlyfailed&buildid=4197441

doutriaux1:
@chaosphere2112 , according to @mcenerney1 diags are using Multi

chaosphere2112:
Going to open a PR for this, though it's not quite ready for release. Changed milestone to 2.4.1.

doutriaux1:
@chaosphere2112 I changed the milestone to 2.6 if it makes it to 2.4.1 ok, but 2.4.1 is a uvcmetrics targeted release.

chaosphere2112:
@doutriaux1 :+1:

chaosphere2112:
@doutriaux1 @aashish24 Since this is a pretty minor issue, I'm going to backburner it for now while I get my GUI on. I'll come back to the PR in a month.

aashish24:
Roger that @chaosphere2112

danlipsa:
@chaosphere2112 This seems to work in my tests. Can you try it with the master branch . I fixed something that sounds like your problem in
https://github.com/UV-CDAT/uvcdat/pull/1896

aashish24:
[...]

PR #1796, opened 2016-01-20
Fix wrong ratio on init

chaosphere2112:
PR for #1795 ; still fails on Diagnostics tests.

The diagnostics tests fail because:

1. diagnostics calls `canvas.portrait()`
2. `canvas.portrait()` calls `backend.portrait()` with height/width from updated `canvasinfo` function
3. `backend.portrait()` calls `backend.resize_or_rotate_window` with height/width
4. `backend.resize_or_rotate_window` sets `bgX` and `bgY` with the updated height/width
5. :beetle: 

I'm probably going to try updating `canvas.portrait` to just update the `size` attribute on the canvas if height/width was not specified; that will make all of the output functions correctly flip the dimensions when needed, and prevent the vicious chain of bugs.

### DIFF
--- a/Packages/vcs/Lib/Canvas.py
+++ b/Packages/vcs/Lib/Canvas.py
@@ -359,8 +359,8 @@ class Canvas(object):
         'ParameterChanged',
         'colormap',
         'backgroundcolor',
-        'bgX',
-        'bgY',
+        '_bgX',
+        '_bgY',
         'display_names',
         '_dotdir',
         '_dotdirenv',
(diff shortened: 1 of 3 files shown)


chaosphere2112:
@danlipsa

aashish24:
@danlipsa are you going to review this branch?

danlipsa:
@aashish24 I have a fix to a similar or the same problem on my branch. Based on the comment at the beginning this seems to be work in progress that is not ready to be merged.

aashish24:
> @aashish24 I have a fix to a similar or the same problem on my branch. Based on the comment at the beginning this seems to be work in progress that is not ready to be merged.

Okay, thanks. In that case, let's close this PR. If there is any mis-understanding, @chaosphere2112 will open it again.

PR #1797, opened 2016-01-20
Adding a vagrant configuration for building with a GUI

jbeezley:
I was having difficulty getting the GUI to build consistently so I did some devops.  See the added [README](https://github.com/UV-CDAT/uvcdat/blob/1a2c75fc9b3b803f9a328c57885d5d2802d04dff/deploy/README.md).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ MAINTAINER UV-CDAT Developers <uvcdat-support@llnl.gov>
 
 RUN mkdir -p /tmp/uvcdat-build
 ADD . /usr/src/uvcdat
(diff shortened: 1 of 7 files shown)


doutriaux1:
@jbeezley @aashish24 I don't know vagrant so I can't really review this. But the changes look decent. @chaosphere2112 @mattben can you please take a look as well? Thanks.

jbeezley:
The issue I was having with ESGF authentication was that curl is [no longer built](https://github.com/UV-CDAT/uvcdat/blob/master/CMakeLists.txt#L452) within UV-CDAT.  Getting ESGF to work is rather sensitive because you need the correct ssl library as well as curl development libraries linked against that ssl.  Adding `libcurl3-openssl-dev` to the provisioning fixed it in this case.

PR #1800, opened 2016-01-21
Disable projected click, fix time values

chaosphere2112:
When clicking on data that was projected, the wrong value was displayed. The issue is that we are not "de-projecting" the coordinates of the mouse. I wasn't able to come up with a quick fix for that (as my ability to reason through projection stuff is pretty limited), so I disabled the value checker for all nonlinear projections. I also removed some redundant logic for extracting the world coordinate (we had a call in `vcs.utils` for this). As an added bonus, we now get correct values when clicking on the plot after having animated it or advanced the current frame beyond the initial frame now. I'm pretty sure the time logic I used is solid, and it should fall back nicely to the "N/A" if it's not able to determine which axis is time.

@sankhesh @doutriaux1 @aashish24 @danlipsa Please review and merge so we can release.

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -162,8 +162,7 @@ class VTKVCSBackend(object):
                     except:
                         st += "Var: %s\nX = %g\nY[%i] = %g\nValue: %g" % (
                             d.array[0].id, X, I, Y, V)
-                except Exception as e:
-                    print e
+                except:
                     st += "Var: %s\nX=%g\nY=%g\nValue = N/A" % (
                         d.array[0].id, X, Y)
         if st == "":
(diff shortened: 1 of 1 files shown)


doutriaux1:
@chaosphere2112 we should really `unproject` we have all the info via the vtk arguments it's not too hard to revert projection in vtk. We should do this the right way.

aashish24:
@chaosphere2112  okay if @danlipsa and @aashish take over this branch?

chaosphere2112:
@aashish24 @doutriaux1 Sure thing, I was just getting this pushed out. I do need to make one tweak to the time axis logic still, but other than that anyone can take it over.

aashish24:
@chaosphere2112 roger that.

aashish24:
@danlipsa can you look in this for the projected plots?

danlipsa:
@doutriaux1 @chaosphere2112 @aashish24 I think a proper fix for this may need plotting_dataset_bounds which I return from plot in my branch: 
https://github.com/UV-CDAT/uvcdat/pull/1806
Probably I could do something now, but I will need to revisit it after that branch is merged.
Given that this never worked, I would say to merge it as is (with projected datasets not showing the coordinates at all) instead holding the release.

aashish24:
@danlipsa this is not holding the release (at least not yet :))

danlipsa:
That's good. :-)

PR #1802, opened 2016-01-25
Update readme master

aashish24:


doutriaux1:
The repo size is broken. We should take it out.

Issue #1812, opened 2016-02-03
Fix the taylor diagram 


aashish24:
Ref: #1806

danlipsa:
Merged: https://github.com/UV-CDAT/uvcdat/pull/1819

PR #1819, opened 2016-02-04
BUG #1812: Fix taylor cutoff.

danlipsa:


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/106

danlipsa:
@charlesdoutriaux @aashish24 Please review.

Packages/vcs/Lib/taylor.py
```diff
@@ -922,15 +922,15 @@ def drawSkill(self, canvas, values, function=None):
         if function is None:
             return
         color = self.skillColor
-        a = MV2.ones((self.detail, self.detail), typecode=MV2.float)
+        a = MV2.ones((self.detail + 1, self.detail + 1), typecode=MV2.float)
```
aashish24: why +1 here? any comment would be useful.

danlipsa: I updated the commit comment. Basically, data was generated with a loop over an right open interval and because of that it was missing the right and top part.

Packages/vcs/Lib/vcsvtk/isolinepipeline.py
```diff
@@ -27,6 +27,11 @@ def _updateVTKDataSet(self):
         data = vcs2vtk.numpy_to_vtk_wrapper(self._data1.filled(0.).flat,
                                             deep=False)
         self._vtkDataSet.GetPointData().SetScalars(data)
+        if (vcs2vtk._DEBUG_VTK):
+            writer = vtk.vtkXMLDataSetWriter()
+            writer.SetFileName("isoline-p.vts")
```
aashish24: do we want this in production code?

Packages/vcs/Lib/taylor.py
aashish24: @danlipsa I think adding a code comment would have been more helpful

Packages/vcs/Lib/vcsvtk/isolinepipeline.py
danlipsa: I don't have a problem with that. We just have to make sure _DEBUG_VTK is not True otherwise a bunch of files are going to be created.  I am open to making this better, though. Production code = many bugs we know about and probably many more we don't know about. :-) So we should make the code as easier/faster to debug as possible. This kind of hooks are very useful and to me it does not make sense to remove them every time I commit something.

Packages/vcs/Lib/taylor.py
danlipsa: See the comment in the new commit.

Packages/vcs/Lib/vcsvtk/isolinepipeline.py
aashish24: @danlipsa I would suggest we take it out. in the past we have discussed leaving the debug code and I think I can understand that it makes it easy for developers to leave the debug code, it certainly add more lines and probably more maintenance to the code in the long run. Also, this will start other developers to follow the same and then we will have a lot of debug code. For example, we don't have this kind of code in VTK or ParaView. 

So I suggest we remove it now and discuss in more detail in the next meeting if that sounds reasonable to you.

Packages/vcs/Lib/taylor.py
aashish24: thanks @danlipsa for adding this!

aashish24:
@danlipsa looks great except the debug part which we should discuss in the next meeting. LGTM :+1: 
- Aashish

danlipsa:
@aashish24 I do think there is a school of thought that says we should leave debugging output in and even be able to enable it at run time not at compile time. See vtkObject::DebugOn, 
http://www.paraview.org/Wiki/ParaView/Properties_Panel#Debugging
and logging libraries such as log4j.

In the same time I share your concern - to much debug code can interfere with development.
I can remove the code for now while we think how/if we can add something like this in the code.

aashish24:
> @aashish24 I do think there is a school of thought that says we should leave debugging output in and even be able to enable it at run time not at compile time. See vtkObject::DebugOn, 
> http://www.paraview.org/Wiki/ParaView/Properties_Panel#Debugging
> and logging libraries such as log4j.
> 
> In the same time I share your concern - to much debug code can interfere with development.
> I can remove the code for now while we think how/if we can add something like this in the code.

+1 thanks @danlipsa. Yes lets discuss it next time. I am all for supporting  easier debugging as much as possible.

aashish24:
LGTM :+1: since dashboards are passing and it's a minor change, I am merging it.

PR #1831, opened 2016-02-16
Fix 1825

chaosphere2112:
Fixes the issue discovered by Ram on the mailing list, where you could not set the missing color for a graphics method using a color's name. Adds a test to cover that usecase.

### DIFF
--- a/Packages/vcs/Lib/utils.py
+++ b/Packages/vcs/Lib/utils.py
@@ -1716,6 +1716,37 @@ def getworldcoordinates(gm, X, Y):
     return wc
 
 
+def rgba_color(color, colormap):
+    """Try all of the various syntaxes of colors and return 0-100 RGBA values."""
+    try:
+        # Is it a colormap index?
+        return colormap.index[color]
+    except ValueError:
+        # Is it a color tuple?
+        if len(color) == 3 or len(color) == 4:
+            for c in color:
+                try:
+                    int(c)
+                except:
+                    break
+            else:
+                if any((c > 100 for c in color)):
+                    r, g, b = (c / 2.55 for c in color[0:3])
+                    if len(color) == 4:
+                        a = color[-1] / 2.55
+                    else:
+                        a = 100
+                else:
+                    r, g, b = color[:3]
+                    if len(color) == 4:
+                        a = color[-1]
+                    else:
+                        a = 100
+                return [r, g, b, a]
+    r, g, b = genutil.colors.str2rgb(color)
+    return [r / 2.55, g / 2.55, b / 2.55, 100]
+
+
 def png_read_metadata(path):
     reader = vtk.vtkPNGReader()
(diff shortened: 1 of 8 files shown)


chaosphere2112:
@doutriaux1 @aashish24

Packages/vcs/Lib/utils.py
```diff
@@ -1716,6 +1716,37 @@ def getworldcoordinates(gm, X, Y):
     return wc
 
 
+def rgba_color(color, colormap):
+    """Try all of the various syntaxes of colors and return 0-100 RGBA values."""
+    try:
+        # Is it a colormap index?
+        return colormap.index[color]
+    except ValueError:
```
aashish24: should be  a IndexError?

Packages/vcs/Lib/utils.py
```diff
@@ -1716,6 +1716,37 @@ def getworldcoordinates(gm, X, Y):
     return wc
 
 
+def rgba_color(color, colormap):
```
aashish24: Could you please add some comments on this code behavior?

Packages/vcs/Lib/utils.py
```diff
@@ -1716,6 +1716,37 @@ def getworldcoordinates(gm, X, Y):
...
+        return colormap.index[color]
+    except ValueError:
+        # Is it a color tuple?
+        if len(color) == 3 or len(color) == 4:
+            for c in color:
+                try:
+                    int(c)
+                except:
+                    break
+            else:
```
aashish24: is it just me or this else is not aligned?

chaosphere2112: No, the `else` is for the `for` loop; it executes if the code doesn't `break`.

Packages/vcs/Lib/utils.py
chaosphere2112: Nope, it's a `ValueError`. We implement `__setitem__` and `__getitem__` on the colormap object, and raise a `ValueError` if the key isn't a number between 0 and 256.

Packages/vcs/Lib/utils.py
aashish24: I see. Using try and except like this is consider best practice?

Packages/vcs/Lib/utils.py
aashish24: Interesting, I would expect a IndexError from this code but your code looks go to me.

aashish24:
LGTM except that I have question on best practices for try and except. 

+2

Packages/vcs/Lib/utils.py
chaosphere2112: Yup, exceptions are cheap in python. [Ask for forgiveness, not permission](http://stackoverflow.com/a/12265860/725871)

Issue #1849, opened 2016-02-20
world coordinate do not respect datawc_x1/x2 


doutriaux1:
@danlipsa we need urgently to fix the datawc_x1/x2/y1/y2 issue this affects I isolated the issue (in meshfill at least which appear in: https://github.com/PCMDI/pcmdi_metrics/issues/296 ) to be coming from: https://github.com/UV-CDAT/uvcdat/blob/master/Packages/vcs/Lib/vcsvtk/meshfillpipeline.py#L53

Looking at the issue above when runnig it with the EXACT same template and datawc_x1/x2 4 times but slightly different "mesh" coordinatesI get:
```
MESH: -0.5 15.5 0.0 9.5
MESH: -0.5 15.0 -0.5 9.5
MESH: -0.5 15.5 -0.5 9.0
MESH: 0.0 15.5 -0.5 9.5
```
which leads to the ugglyness in the issues (ticks offsets, triangle on top of each others)

Another example would be:

```python
import cdms2
import vcs
f=cdms2.open(vcs.sample_data+"/sampleCurveGrid4.nc")
s=f("sample")
x=vcs.init()
m=x.createmeshfill()
m.datawc_x1 = -20
m.datawc_x2 = 20
m.datawc_y1 = -20
m.datawc_y2 = 20
x.plot(s,m)
x.png("bad")
```

png, far from being the requested bounds

![bad](https://cloud.githubusercontent.com/assets/2781425/13193372/06eaa826-d72c-11e5-91e9-699b7dd20f22.png)

danlipsa:
@doutriaux1 Could I get the original script/data from https://github.com/PCMDI/pcmdi_metrics/issues/296? I have seen and solved something similar for plots of boxfill + vector. There, the bounds for boxfill data were slightly larger than the bounds for vector the vector which resulted in similar duplication of labels.

danlipsa:
Does this fix the issue?
https://github.com/UV-CDAT/uvcdat/pull/1855

PR #1855, opened 2016-02-23
BUG #1849: Re-enable datawc for linear projection.

danlipsa:
For datasets using a geographic projection, datawc
is only used to specify wrapping (translation of the origin,
for instance from 0:360 to -180:180) and flipping.

### DIFF
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -48,8 +48,8 @@ class VTKVCSBackend(object):
             'renderer',
             'vtk_backend_grid',
             'dataset_bounds',
-            # the same as dataset_bounds but centered on the same value and in the
-            # same order as gm bounds. See vcs2vtk.getBoundsForPlotting
+            # the same as vcs.utils.getworldcoordinates for now. getworldcoordinates uses
+            # gm.datawc_... or, if that is not set, it uses data axis margins (without bounds).
             'plotting_dataset_bounds',
             # this may be slightly smaller than the data viewport. It is used
             # for vectorpipeline when drawn on top of a boxfill for instance
@@ -690,14 +690,13 @@ class VTKVCSBackend(object):
             if hasattr(plot, 'onClosing'):
                 plot.onClosing(cell)
 
-    def plotContinents(self, x1, x2, y1, y2, projection, wrap, vp, priority, **kargs):
+    def plotContinents(self, wc, projection, wrap, vp, priority, **kargs):
         contData = vcs2vtk.prepContinents(self.canvas._continentspath())
         contMapper = vtk.vtkPolyDataMapper()
         contMapper.SetInputData(contData)
         contActor = vtk.vtkActor()
         contActor.SetMapper(contMapper)
-        contActor = vcs2vtk.doWrap(
-            contActor, [x1, x2, y1, y2], fastClip=False)
+        contActor = vcs2vtk.doWrap(contActor, wc, fastClip=False)
         if vcs2vtk._DEBUG_VTK:
             writer = vtk.vtkXMLPolyDataWriter()
             writer.SetFileName("cont.vtp")
@@ -709,7 +708,7 @@ class VTKVCSBackend(object):
             cpts = contData.GetPoints()
             # we use plotting coordinates for doing the projection so
             # that parameters such that central meridian are set correctly.
-            geo, gcpts = vcs2vtk.project(cpts, projection, [x1, x2, y1, y2])
+            geo, gcpts = vcs2vtk.project(cpts, projection, wc)
             contData.SetPoints(gcpts)
 
             if vcs2vtk._DEBUG_VTK:
@@ -754,7 +753,7 @@ class VTKVCSBackend(object):
         vtk_backend_grid = kargs.get("vtk_backend_grid", None)
         self.fitToViewportBounds(contActor,
                                  vp,
-                                 wc=[x1, x2, y1, y2], geo=geo,
+                                 wc=wc, geo=geo,
                                  geoBounds=vtk_backend_grid.GetBounds(),
                                  priority=priority,
                                  create_renderer=True)
@@ -1281,7 +1280,6 @@ class VTKVCSBackend(object):
                     pass
                 if flipX:
                     cam.Azimuth(180.)
-
         T = vtk.vtkTransform()
         T.Scale(xScale, yScale, 1.)
 
--- a/Packages/vcs/Lib/VTKPlots.py
+++ b/Packages/vcs/Lib/VTKPlots.py
@@ -697,11 +697,6 @@ class VTKVCSBackend(object):
         contActor = vtk.vtkActor()
         contActor.SetMapper(contMapper)
         contActor = vcs2vtk.doWrap(contActor, wc, fastClip=False)
-        if vcs2vtk._DEBUG_VTK:
-            writer = vtk.vtkXMLPolyDataWriter()
-            writer.SetFileName("cont.vtp")
-            writer.SetInputData(contActor.GetMapper().GetInput())
-            writer.Write()
 
         if projection.type != "linear":
             contData = contActor.GetMapper().GetInput()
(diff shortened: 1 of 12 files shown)


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/112

aashish24:
@danlipsa why so many baselines changed? It seems to me that we are back to half cell on most of them?

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -690,14 +692,13 @@ def onClosing(self, cell):
             if hasattr(plot, 'onClosing'):
                 plot.onClosing(cell)
 
-    def plotContinents(self, x1, x2, y1, y2, projection, wrap, vp, priority, **kargs):
+    def plotContinents(self, wc, projection, wrap, vp, priority, **kargs):
```
doutriaux1: good idea!

Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -253,10 +253,11 @@ def genGridOnPoints(data1, gm, deep=True, grid=None, geo=None,
...
         bounds = pts.GetBounds()
         xm, xM, ym, yM = [bounds[0], bounds[1], bounds[2], bounds[3]]
-        # we use plotting coordinates for doing the projection
+        # We don't use the zooming feature (gm.datawc) for geographic
+        # projections. We use wrapped coordinates for doing the projection
         # such that parameters such that central meridian are set correctly
-        geo, geopts = project(pts, projection, getBoundsForPlotting(
-            [gm.datawc_x1, gm.datawc_x2, gm.datawc_y1, gm.datawc_y2], [xm, xM, ym, yM], wrap))
+        geo, geopts = project(pts, projection, getWrappedBounds(
+            [gm.datawc_x1, gm.datawc_x2, gm.datawc_y1, gm.datawc_y2], [xm, xM, ym, yM], None))
```
doutriaux1: why wrap None here? Does geo = None means plain array or does it means no projection? I can't remember for sure.

Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -1911,6 +1914,20 @@ def vtkIterate(iterator):
...
+    if geo:
+        return [x1, x2, y1, y2]
+    assert (x1 < x2 and y1 < y2)
+    if not numpy.allclose([x1gm, x2gm], 1.e20):
+        x1, x2 = [x1gm, x2gm]
+    if (isinstance(y1gm, numbers.Number) and not numpy.allclose([y1gm, y2gm], 1.e20)):
+        y1, y2 = [y1gm, y2gm]
+    return [x1, x2, y1, y2]
+
+
```
doutriaux1: isn't it essentially what vcs.utils.getworldcoordinate does?

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
```diff
@@ -114,7 +114,8 @@ def _plotInternal(self):
         if self._maskedDataMapper is not None:
             self._mappers.insert(0, self._maskedDataMapper)
 
-        x1, x2, y1, y2 = self.getBoundsForPlotting()
+        plotting_dataset_bounds = self.getPlottingBounds()
+        x1, x2, y1, y2 = plotting_dataset_bounds
```
doutriaux1: do we still need x1,x2,y1,y2?

doutriaux1:
@danlipsa looks good, let's add a quick test to make sure we don' break this again.

danlipsa:
@aashish24 @doutriaux1 It is true that we are back to half cell - this is what vcs.utils.getworldcoordinate tells us to do - to zoom into the dataset.

We can try to get rid of vcs.utils.getworldcoordinate and only follow datawc. getworldcoordinates sets zooming based on the datawc or if that is not set it sets it based on the data (without bounds).

Note there is still no cutting or misalignment so we did not loose that.

doutriaux1:
@aashish24 where do you see a half cell? worldcoordinate uses your  axes values, we can try to tweak it to use the axes bounds rather than the just the axes. I'm opened to that.

Packages/vcs/Lib/vcs2vtk.py
danlipsa: @doutriaux1 @aashish24 It ends up being the same thing. Maybe instead of using getworldcoordinates which sets zooming to be datawc or the data without bounds, we should use datawc or the full data - with bounds. What do you guys think? That might cause some incompatibilities as it changes how the package behaves.

Packages/vcs/Lib/vcsvtk/boxfillpipeline.py
danlipsa: Yes, they are still used a little lower.

danlipsa:
@doutriaux1 Look for instance at https://github.com/UV-CDAT/uvcdat-testdata/pull/112/files. Instead of showing the whole dataset we zoom in and cut the cells on the margine in half.

Packages/vcs/Lib/vcs2vtk.py
doutriaux1: i'd say it's worth trying at least.

doutriaux1:
Ok I found a few ones finally. Thanks. Let's try to have getworldcoordinate return the bounds. Historically it use to be just the axes and originally I wanted to preserve this, to help users feel more confident with the VTK transition. At this point I don't think we need to do this anymore. Let's have getworldcoordinate return the real world coordinates of the data, i..e with axes bounds.

danlipsa:
@doutriaux1 @aashish24 The problem with modifying getworldcoordinates is that coordinates depend on the type of plot we have: boxfill uses bounds, isofill, isoline and vector do not. 
Question: Why do we create both cell data and point data datasets? This does not seem to be specified in the nc file.
Regardless, it is possible for different variables to have different bounds coming from an nc file in which case there is nothing we can do.

I already started passing backend.dataset_bounds using **kargs mechanism. I would use that instead and maybe get rid of getworldcoordinates alltogether as some point. Maybe we do this in two steps. 1. Try to show the full dataset if datawc is not set. I will use dataset_bounds for this 2. get rid of getworldcoordinates all together as a separate issue.

doutriaux1:
@danlipsa the worldcoordinates for all graphic method should come from bounds, in the case of meshfill there are no axes so use data.getMashfill().min/max() 
boxfill/meshfill uses celldata because we draw cells/polygones defined by the data itself, isolines and isofill only need point data since we interpolate. Does that make sense?

danlipsa:
@doutriaux1 I was thinking that always using cell data would make the code more consistent and simplify it. If we need point data in certain plots we can easily convert cell data to point data. So, we always create the data that is specified in the file, regardless of what a specific plot needs. Note that vtkCellDataToPointData works by averaging cells used by every point, which will give you the same bounds as for the cell data which is a nice thing to have when doing several plots in the same image.

doutriaux1:
@danlipsa that's fine with me. I thought isofill was requiring point data (filter) if it doesn't then it's ok with me. Although point data is nice for unstructured grid since we don't always have the whole cell information. Anyhow let's do this in a separate PR.

danlipsa:
@doutriaux1 Definitely a separate PR.

Packages/vcs/Lib/vcs2vtk.py
danlipsa: This is a bug. Thanks!

danlipsa:
@aashish24 @doutriaux1 Note there are missing triangles when zooming into mesh data:
vcs_test_meshfill_zoom
Probably has to do with wrapping of meshfill data, but I don't know why it happens. I will look into it.

danlipsa:
@doutriaux1 besides that problem it should be ready to merge. I think it is worth trying to see if it fixes https://github.com/PCMDI/pcmdi_metrics/issues/296. @aashish24 We decided to leave the half cells in as this was the original behavior of uvcdat and we'll be faster in fixing the issue at hand. We can enable showing the whole dataset as a separate PR.

doutriaux1:
@danlipsa yes! It does fix this issue:
test_4seasrms

danlipsa:
@doutriaux1 Great!

danlipsa:
@doutriaux1 Interesting plot by the way.

danlipsa:
@doutriaux1 What does the plot mean? For each cell, it shows 4 things? What are those? I am curious. :-)

doutriaux1:
Each cell is basically a score skill for a model/variable. The cell itself is broken into N (4 in this case) references. That is to make sure your score does not depend on the set set of observation you used.

durack1:
@danlipsa, @gleckler1 will be delighted that you think the `gleckler plot` is "interesting".. He's going to be making many more of them in the coming months..

danlipsa:
@durack1 Indeed, I think the plot looks good - it draw me in even without knowing what it means.

PR #1864, opened 2016-03-02
Updated VTK to use current uvcdat-master

aashish24:
uvcdat-master is synced up with VTK/master

### DIFF
--- a/CMake/cdat_modules/vtk_pkg.cmake
+++ b/CMake/cdat_modules/vtk_pkg.cmake
@@ -1,4 +1,4 @@
 set(VTK_SOURCE ${GIT_PROTOCOL}github.com/UV-CDAT/VTK.git )
 set(VTK_MD5)
-set(VTK_BRANCH uvcdat-2.4.0)
+set(VTK_BRANCH uvcdat-master)
 add_cdat_package_dependent(VTK "" "" ON "CDAT_BUILD_GRAPHICS" OFF)


aashish24:
@danlipsa you can merge your branches into this branch.

danlipsa:
@aashish24 Do you need to update uvcdat-master to point to the latest VTK or should I do that?

doutriaux1:
@aashish24 , @danlipsa is right uvcdat-master is 22 commits ahead, 2295 commits behind Kitware:master.

doutriaux1:
@aashish24 looks like https://github.com/UV-CDAT/uvcdat/compare/update_vtk merged into `master` not `uvcdat-master`, should we use `master` from now on, sounds like a better idea no?

doutriaux1:
actually master is 3645 commits behind Kitware:master so I'm lost I'm not sure what you did @aashish24

aashish24:
@doutriaux1 @danlipsa sorry since I didn't force push the uvcdat-master, when my script pushed it, it got rejected and I didn't see that. the uvcdat-master in UVCDAT/vtk should be on correct sha now. 

The reason, we don't want to use master because we want consistency across all repos (use uvcdat-master) or else if other repo's master get's updated things may break.

aashish24:
@doutriaux1 @danlipsa this should be ready for review.

doutriaux1:
thanks @aashish24

aashish24:
@doutriaux1 I updated (rebased) this branch to trigger rebuild.

danlipsa:
@aashish24 Would it make sense to use a certain SHA1 instead of uvcdat-master? This way you could go back to a previous uvcdat SHA-1 and have reproducible behavior. With current behavior you always use uvcdat-master, whatever that may be at the moment.

doutriaux1:
@danlipsa we only use uvcdat-master for in-between releases, when we release we point to a sha1or more exactly to a tag point (same thing anyway)

danlipsa:
I'll update VTK in the update proj4 PR, so I'll close this one.

danlipsa:
@doutriaux1 I agree that we have reproducibility for releases because we tag external modules such as VTK. However in between releases we don't as, for VTK, we always point to the latest VTK master. This is something that we may want to add in the future. I don't think it will cost us anything. We just point to a certain SHA1 instead of the top of a branch.

PR #1870, opened 2016-03-03
Mintics now properly support lists

chaosphere2112:
I tried to do this in #1869, but @doutriaux1 beat me to the punch and merged it before I ran into this :smile: 

Miniticks were not fetching lists from `vcs.elements`, this has been remedied. Updated the test... my push to the baseline change was merged already.

### DIFF
--- a/Packages/vcs/Lib/template.py
+++ b/Packages/vcs/Lib/template.py
@@ -1176,13 +1176,16 @@ class P(object):
                         txs.append(wc[0])
                     tstring.append(loc[l])
         # now does the mini ticks
-        if getattr(gm, axis + 'mtics' + number) != '':
+        mintics = getattr(gm, axis + 'mtics' + number)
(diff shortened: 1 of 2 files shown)


Packages/vcs/Lib/template.py
```diff
@@ -1176,13 +1176,16 @@ def drawTicks(self, slab, gm, x, axis, number,
                         txs.append(wc[0])
                     tstring.append(loc[l])
         # now does the mini ticks
-        if getattr(gm, axis + 'mtics' + number) != '':
+        mintics = getattr(gm, axis + 'mtics' + number)
+        if mintics != '':
+            if isinstance(mintics, str):
+                mintics = vcs.elements["list"][mintics]
```
doutriaux1: @chaosphere2112 can it ever be '*' ?

Issue #1811, opened 2016-02-03
Fix projected click


aashish24:
Once #1806 goes into master.

aashish24:
Ref: #1800

danlipsa:
This was partially fixed by
https://github.com/UV-CDAT/uvcdat-testdata/pull/117/commits
We still have to implement this for point datasets.

danlipsa:
This implements point info for point datasets.
https://github.com/UV-CDAT/uvcdat/pull/1966

PR #1878, opened 2016-03-15
BUG #1811: Show point information for plots using geographic projecti…

danlipsa:
…ons.


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/117

danlipsa:
@aashish24 @doutriaux1 Please review. Picking works for boxfill and meshfill. We can make it work for all plots after https://github.com/UV-CDAT/uvcdat/issues/1881

aashish24:
@danlipsa why would this baseline changed? test_vcs_settings_color_name_rgba_meshfill.png

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -131,47 +140,97 @@ def leftButtonPressEvent(self, obj, event):
...
-                X = (x - t.data.x1) / (t.data.x2 - t.data.x1) * (x2 - x1) + x1
-                Y = (y - t.data.y1) / (t.data.y2 - t.data.y1) * (y2 - y1) + y1
-
-                # Ok we now have the X/Y values we need to figure out the
-                # indices
-                try:
-                    I = d.array[0].getAxis(-1).mapInterval((X, X, 'cob'))[0]
+            # for non-linear projection or for meshfill. Meshfill is wrapped at
+            # VTK level, so vcs calculations do not work.
+            if gm.projection != "linear" or gm.g_name == 'Gfm':
```
aashish24: @danlipsa why not use the same code (under if) for both cases?

Packages/vcs/Lib/VTKPlots.py
```diff
@@ -131,47 +140,97 @@ def leftButtonPressEvent(self, obj, event):
...
+                selector = vtk.vtkHardwareSelector()
+                datasetRenderer = d.backend['dataset_renderer']
+                continentsRenderer = d.backend.get('continents_renderer')
+                dataset = d.backend['vtk_backend_grid']
+                if (datasetRenderer and dataset):
+                    selector.SetRenderer(datasetRenderer)
+                    selector.SetArea(xy[0], xy[1], xy[0], xy[1])
+                    selector.SetFieldAssociation(vtk.vtkDataObject.FIELD_ASSOCIATION_CELLS)
+                    # We want to be able see information behind continents
```
aashish24: @danlipsa why field association?

danlipsa:
@aashish24 I think has to do with the fact that now I use one renderer for all mappers. Meshfill and boxfill use several mappers one for each range of scalar values. I assume using only one renderer changed the order in which those mappers are rendered. We can file an issue fix this many mappers problem.

Packages/vcs/Lib/vcs2vtk.py
```diff
@@ -507,6 +507,10 @@ def genGrid(data1, data2, gm, deep=True, grid=None, geo=None):
         vg.SetPoints(geopts)
     else:
         vg = grid
+    # Add a GlobalIds array to keep track of cell ids throughout the pipeline
+    globalIds = numpy_to_vtk_wrapper(numpy.arange(0, vg.GetNumberOfCells()), deep=True)
```
aashish24: @danlipsa why we have to do this? Meaning why we have to explicitly append the global ids?

Packages/vcs/Lib/VTKPlots.py
danlipsa: Because this code does not work now for isofill, isoline and vectors. When we fix dataset creation we can get rid of the vcs code.

Packages/vcs/Lib/VTKPlots.py
danlipsa: Because we want cellids which will give us cell scalars.

aashish24: Ah, I see thanks. So are we not looking for point scalars in this case?

Packages/vcs/Lib/VTKPlots.py
aashish24: @danlipsa thanks. And I guess we have an issue for it?

Packages/vcs/Lib/vcs2vtk.py
danlipsa: Because the hardware selector renders and gets the cell ids of the geometry which are not usually the same as the cell ids of the dataset.

aashish24:
> @aashish24 I think has to do with the fact that now I use one renderer for all mappers. Meshfill and boxfill use several mappers one for each range of scalar values. I assume using only one renderer changed the order in which those mappers are rendered. We can file an issue fix this many mappers problem.

Sure, if you can create an issue for it (if it does not exists already), let met know.

Packages/vcs/Lib/VTKPlots.py
danlipsa: That is correct. For point data, the user has to click fairly close the a point to see the value. I have to experiment with that, I am not sure.

Packages/vcs/Lib/VTKPlots.py
danlipsa: Yes we do:
https://github.com/UV-CDAT/uvcdat/issues/1881

Packages/vcs/Lib/vcs2vtk.py
aashish24: > Because the hardware selector renders and gets the cell ids of the geometry which are not usually the same as the cell ids of the dataset.

this is interesting. But by default should they be same?

Packages/vcs/Lib/VTKPlots.py
aashish24: thanks @danlipsa

aashish24:
@danlipsa other than some of the questions I have, this branch looks good to me. I did manually compiled the branch and use the GUI for testing. Looks good to me :+1:

Packages/vcs/Lib/VTKPlots.py
aashish24: > That is correct. For point data, the user has to click fairly close the a point to see the value. I have to experiment with that, I am not sure

Do we have test for it?

Packages/vcs/Lib/vcs2vtk.py
danlipsa: What do you mean 'by default'? In the uvcdat case we have threshold + different mapper for different scalar ranges so the ids are not the dataset ids. If you have only the dataset in the pipeline I think only the polydata will not change ids. Maybe there are others but I would not bet on it :-)

aashish24: > If you have only the dataset in the pipeline I think only the polydata will not change ids. Maybe there are others but I would not bet on it :-)

I see. yeah, I won't either.

aashish24:
@danlipsa the only thing I am not clear about is point data scalars other than that I think you covered it pretty well.

danlipsa:
So, for point data, the hardware selector renders the points using pointids instead of color. So we can get the pointid and from it we can get the point scalar. The only question I have is how close do you have to click on the point to get the point id and how user friendly is that. We might be able to set some sort of point size/range so that you don't have to click exactly on the point to get its value.

danlipsa:
> Sure, if you can create an issue for it (if it does not exists already), that would be great.
See:
https://github.com/UV-CDAT/uvcdat/issues/1884

Packages/vcs/Lib/VTKPlots.py
danlipsa: This is not implemented yet. I'll do it together with or after
https://github.com/UV-CDAT/uvcdat/issues/1881

aashish24:
> Sure, if you can create an issue for it (if it does not exists already), that would be great.
> See:
> #1884

thanks @danlipsa

Packages/vcs/Lib/VTKPlots.py
aashish24: > This is not implemented yet. I'll do it together with or after
> #1881

thanks. Should we file an issue for it @danlipsa ?

aashish24:
> So, for point data, the hardware selector renders the points using pointids instead of color. So we can get the pointid and from it we can get the point scalar. The only question I have is how close do you have to click on the point to get the point id and how user friendly is that. We might be able to set some sort of point size/range so that you don't have to click exactly on the point to get its value.

Sure. I guess we should create an issue for it as well.

Packages/vcs/Lib/VTKPlots.py
danlipsa: @aashish24 See https://github.com/UV-CDAT/uvcdat/issues/1885

danlipsa:
@doutriaux1 I think this is ready to go. :-)

aashish24:
Ping @charlesdoutriaux

aashish24:
@doutriaux1 please approve.

doutriaux1:
just finished building. Testing the click and approving

doutriaux1:
@danlipsa so it does not work on isofill. Correct

aashish24:
@doutriaux1 right, it does not workon isofill (https://github.com/UV-CDAT/uvcdat/issues/1885)

doutriaux1:
Ok approving this for now then.

danlipsa:
@doutriaux1 Make sure you merge the data branch as well: UV-CDAT/uvcdat-testdata#117

PR #1893, opened 2016-03-24
Convert VCS to use setuptools instead of distutils

chaosphere2112:
Converted `vcs` to use the `setuptools` package instead of `distutils`, which gives us access to the wonderful `python setup.py develop` option, which keeps your package in sync with your source code (eliminating a whole host of repetitive calls to `python setup.py install`). Also fixes some stray unicode characters that I most likely stuck into the code at some point. My bad.


aashish24:
Nice :clap: :clap:  my only desire was to have in the huboard first.

aashish24:
LGTM :+1: changing vcs/Lib --> vcs/vcs  is very nice too.

Packages/vcs/vcs/dv3d.py
```diff
@@ -156,8 +156,13 @@ def getConfigurationState(self, pname, **args):
         return self.cfgManager.getConfigurationState(pname, **args)
 
     def add_property(self, name):
-        fget = lambda self: self.getParameter(name)
-        fset = lambda self, value: self.setParameter(name, value)
+
+        def fget(self):
+            return self.getParameter(name)
```
aashish24: @chaosphere2112 what this has to do with setup tools?

Packages/vcs/vcs/queries.py
```diff
@@ -431,8 +431,7 @@ def is3d_scalar(obj):
    ...
 
 """
-    if (isinstance(obj, dv3d.Gf3Dscalar)
-            or isinstance(obj, dv3d.Gf3DDualScalar)):
+    if (isinstance(obj, dv3d.Gf3Dscalar) or isinstance(obj, dv3d.Gf3DDualScalar)):
```
doutriaux1: could probably rewrite this to 
```
isinstance(obj,(dv3d.Gf3Dscalar, dv3d.Gf3DDualScalar)):
```

doutriaux1:
@chaosphere2112 do you mind if we merge this post 2.4.1 ?

Packages/vcs/vcs/dv3d.py
chaosphere2112: I had to update the flake8 test because I changed the directory path of the package, and then to make sure the update to the test worked, I had to run it. This uncovered a number of flake8 failures (due to some updates to flake8 that I guess we merged recently?).

chaosphere2112:
@doutriaux1 Yeah, that's why I milestoned it at 2.6

doutriaux1:
@chaosphere2112 I needed this right now, so I changed my mind merged now. 2.4.1 takes too long to come out anyway.

Issue #1894, opened 2016-03-25
Boxfill legend incorrect when there are fewer colors than levels


chaosphere2112:
When you plot a boxfill and use the `color_1` and `color_2` attributes (but leave `level_1` and `level_2` alone), the legend doubles up for the last level color. Also, the labeling on the legend is excessive (we probably don't need 10+ labels when there are only 4 levels).

```python
import vcs, cdms2
f = cdms2.open(vcs.sample_data + "/clt.nc")
s = f("clt")
x = vcs.init()
box = vcs.createboxfill()
# Set them to the obnoxious end of the colormap so we can see the steps between each color
box.color_1 = 242
box.color_2 = 246
x.plot(s, box)
```

results in:

![out](https://cloud.githubusercontent.com/assets/718512/14052964/7eeb0828-f28b-11e5-9ac0-67036cd0c2cb.png)

I've got a PR incoming to fix this.

doutriaux1:
@chaosphere2112 ia that finished?

chaosphere2112:
@doutriaux1 I've been waiting for feedback from you and @aashish24 for 2 weeks

PR #1895, opened 2016-03-25
Fixes doubled colors in legend and reduces label count

chaosphere2112:
Fixes the issues outlined in #1894; still need to add a test.


aashish24:
@chaosphere2112 when you add a test could you please reopen this again?

aashish24:
@chaosphere2112 can you resolve the merge conflicts?

chaosphere2112:
@aashish24 Yeah; there's another bug that I ran into that I'm running down right now. I'll get an update for this by EOD.

aashish24:
Sure, thanks.

chaosphere2112:
@aashish24 I ran into a few more issues, I'll clean this up in the morning.

chaosphere2112:
@doutriaux1 @aashish24 

[Here's where this PR stands currently on CDASH](https://open.cdash.org/viewTest.php?onlyfailed&buildid=4329263).

So this is starting to get a little bit out of hand. I've done some major improvements to some of the baselines ([test_vcs_boxfill_zero](https://open.cdash.org/testDetails.php?test=440066621&build=4329263) and [test_vcs_boxfill_lev1_lev2](https://open.cdash.org/testDetails.php?test=440078091&build=4329263) are notably better), but then there's... debatable movement on others ([test_vcs_boxfill_lev1_lev2_ext1](https://open.cdash.org/testDetails.php?test=440025816&build=4329263) is a good example of one that isn't *bad*, but it's certainly not *better* than it was). I think I fixed a bug in [test_vcs_issue_960](https://open.cdash.org/testDetails.php?test=440022646&build=4329263) (it looks like the last label was missing from the legend, for some reason). How do people feel about less "pretty" numbers for some scales? I can try and be a little smarter about reverting to the old method (vcs.mkscale), but my first attempt at doing so made `test_vcs_boxfill_lev1_lev2` revert back to the old legend.

aashish24:
@chaosphere2112 less pretty numbers you meant they are decimans now? I like the new legend but I think it would be great to have an option to clamp to round numbers since those are easy to read but in some cases you may not want that.

chaosphere2112:
@aashish24 The users still have the ability to use round numbers; if they don't specify `level_1` and `level_2` on the boxfill, it'll behave exactly the same as it always has (which is why I'm not failing more baselines 😉 ). The "less pretty" numbers are a kind of weird artifact of how I'm generating the labels; looking at the code, I'm guessing there's an off-by-one error somewhere that I've missed which is responsible for an extra label making the levels be shifted off of the "nice" values.

chaosphere2112:
@aashish24 @doutriaux1 Alright, this is ready. I fixed the off-by-one errors, so now the same output should be generated for the same input, plus improvements (like a missing level, and lining up the last level, and other goodies).

chaosphere2112:
@aashish24 ping

aashish24:
thanks @chaosphere2112 will look into it today.

Packages/vcs/vcs/boxfill.py
```diff
@@ -734,6 +735,79 @@ def xyscale(self, xat='linear', yat='linear'):
...
+            high_end = varmax
+        else:
+            low_end = self.level_1
+            high_end = self.level_2
+
+        if self.boxfill_type == "log10":
+            low_end = numpy.ma.log10(low_end)
+            high_end = numpy.ma.log10(high_end)
+
+
```
aashish24: @chaosphere2112 nitpick, do we need this many empty lines?

Packages/vcs/vcs/utils.py
```diff
@@ -994,21 +994,16 @@ def mklabels(vals, output='dict'):
...
             aa = aa * 10.
-        idig = numpy.ma.maximum(
-            idig,
-            numpy.ma.floor(
-                numpy.ma.log10(
-                    aa *
-                    numpy.ma.power(
-                        10.,
-                        idigleft))))
+        idig = numpy.ma.maximum(idig, numpy.ma.floor(numpy.ma.log10(aa * numpy.ma.power(10., idigleft))))
```
aashish24: can we wrap this long line?

Packages/vcs/vcs/vcsvtk/boxfillpipeline.py
```diff
@@ -286,7 +245,7 @@ def _plotInternalBoxfill(self):
 
         # Colortable bit
         # make sure length match
-        numLevels = len(self._contourLevels)
+        numLevels = len(self._contourLevels) - 1
```
aashish24: Why -1 now?

chaosphere2112: It was wrong before. When we use it right below, we pad the colors in `_contourColors` to the length of `numLevels`, but `_contourLevels` represents the bounds of each "level"; this means that there was one more color than there should have been, and led to a duplicated color appearing at the end of the legend (hard to see if you're using lots of levels, but very easy to see when you drop it down to a small number of levels). That's actually the root of this whole pull request, but I hit on some other sticky widgets along the way.

Packages/vcs/vcs/boxfill.py
chaosphere2112: Nope, I'll remove the extra one.

Packages/vcs/vcs/utils.py
chaosphere2112: Our max line length is 120, this line is 106, so, no. This is way more readable than the crazy line wrapping I replaced.

Packages/vcs/vcs/boxfill.py
aashish24: thanks @chaosphere2112

Packages/vcs/vcs/utils.py
aashish24: you are right @doutriaux1 did push us for 120 😄

Packages/vcs/vcs/vcsvtk/boxfillpipeline.py
aashish24: ah I see. Thanks @chaosphere2112 for clarification.

aashish24:
LGTM 👍

Packages/vcs/vcs/utils.py
doutriaux1: exactly for this kind of lines. I'm a visionary @aashish24

aashish24: > exactly for this kind of lines. I'm a visionary @aashish24

@doutriaux1 you are a :star: too :smile:

doutriaux1:
@chaosphere2112 don't forget to delete branch after merging

chaosphere2112:
@doutriaux1 Sorry, got excited and merged when I was going to bed 😄

doutriaux1:
you mean you actually take time to sleep! 😜

PR #1928, opened 2016-04-14
Allow canvas.plot to handle vcs addons

sampsonbryce:
Fixes a bug where passing a vcsaddon graphics method to canvas.plot would raise an exception.


doutriaux1:
you're first commit! Yeah!

Issue #1854, opened 2016-02-23
uvcdat testing checks too many images


danlipsa:
An image match is tested for image.png and for image_*.png.

Instead of image_*.png it should check image_?.png

doutriaux1:
that would probably be better.

aashish24:
:+1:

PR #1943, opened 2016-04-21
ENH #1854: Tests will check for alternate baselines using basename_[1…

danlipsa:
…-9].ext

Previously the pattern used was baseline.*\.ext

### DIFF
--- a/testing/checkimage.py
+++ b/testing/checkimage.py
@@ -9,6 +9,7 @@ import numpy
 import vtk
 import os
 import os.path
+import re
 import sys
 import logging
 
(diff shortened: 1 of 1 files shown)


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/127

aashish24:
LGTM :+1:

danlipsa:
@doutriaux1 Please review.

doutriaux1:
@danlipsa I'm cool with your changes, but in general I think it's unsafe to use alternate baselines.

First it adds work to maintaining the baselines, also I'm always worried a test gets better doesn't need the laternate baseline anymore, then we fix an issue with that test. A few week later a new change breaks the test but we don't see it because the old "bad" baseline is still here and passes.

danlipsa:
@doutriaux1 I am not happy with alternate baselines but I think they are a necessary evil. It boils down to  the following choice: Should we spend a week to debug a small difference in how images are rendered on a specific platform, possibly narrow it down to a OpenGL difference in how the standard is interpreted or a OpenGL bug -- or spend a minute to add a new baseline. VTK uses alternate baselines. We run on many platforms/compilers. I think we only use 3 alternate baselines at most. For uvcdat I think we have at most one.

doutriaux1:
agreed

Issue #1881, opened 2016-03-15
vcs should create a dataset independent of the plot it is used for


danlipsa:
Using the same data (clt variable in clt.nc) different datasets are created for boxfill (this uses cell data) and isofill (this uses point data). This may result in slightly incorrect images.

I propose to always create the dataset that can be created from the data - if bounds are available create cell data otherwise create point data. Then, depending on the plot, we can use pointtocell or celltopoint filters to convert the data in what we need in specific plots.

This would simplify the code as we always use the same function to create a dataset. It will also enable picking for point plots (isofill, isoline, vector).

PR #1948, opened 2016-04-25
ENH #1881: Create the dataset read from the file, regardless of what …

danlipsa:
…the plot needs.

Previously, we created a point dataset if the plot needed one. The point dataset created
was incorrect, as it could use the cell centered values as point values. The grid was also
smaller - it used cell centers as points.

### DIFF
--- a/Packages/cdms2/Lib/grid.py
+++ b/Packages/cdms2/Lib/grid.py
@@ -496,14 +496,6 @@ class AbstractRectGrid(AbstractGrid):
         else:
             latbnds = lat.genGenericBounds()
 
-        # Stretch latitude bounds to +/- 90.0
-        if ascending:
-            latbnds[0,0] = min(latbnds[0,0],-90.0)
-            latbnds[-1,1] = max(latbnds[-1,1],90.0)
-        else:
-            latbnds[0,0] = max(latbnds[0,0],+90.0)
-            latbnds[-1,1] = min(latbnds[-1,1],-90.0)
-
         # Get longitude bounds
         lon = self.getLongitude()
         if len(lon)>1:
--- a/Packages/cdms2/Lib/avariable.py
+++ b/Packages/cdms2/Lib/avariable.py
@@ -363,6 +363,16 @@ class AbstractVariable(CdmsObj, Slab):
             if axisMatches(self.getAxis(i), axis_spec):
                 return i
         return -1
+    
+    def hasCellData(self):
+        '''
+        If any of the variable's axis has explicit bounds, we have cell data
+        otherwise we have point data.
+        '''
+        for axis in self.getAxisList():
+            if (axis.getExplicitBounds() is not None):
+                return True
+        return False
 
     def getAxisListIndex (self, axes=None, omit=None, order=None):
         """Return a list of indices of axis objects;
@@ -510,7 +520,6 @@ class AbstractVariable(CdmsObj, Slab):
                 result = None
                 
         return result
-
 
     # Get an order string, such as "tzyx"
     def getOrder(self, ids=0):
@@ -915,7 +924,9 @@ class AbstractVariable(CdmsObj, Slab):
                 else:
                     delta_beg_wrap_dimvalue = ncyclesrev*cycle
 
-                axis.setBounds(axis.getBounds() - delta_beg_wrap_dimvalue)
+                isGeneric = [False]
+                b = axis.getBounds(isGeneric) - delta_beg_wrap_dimvalue
+                axis.setBounds(b, isGeneric=isGeneric[0])
                 
                 axis[:]= (axis[:] - delta_beg_wrap_dimvalue).astype(axis.typecode())
 
--- a/Packages/cdms2/Lib/axis.py
+++ b/Packages/cdms2/Lib/axis.py
@@ -105,8 +105,8 @@ def getAutoBounds():
     return _autobounds
 
 # Create a transient axis
-def createAxis(data, bounds=None, id=None, copy=0):
-    return TransientAxis(data, bounds, id, copy=copy)
+def createAxis(data, bounds=None, id=None, copy=0, genericBounds=False):
+    return TransientAxis(data, bounds=bounds, id=id, copy=copy, genericBounds=genericBounds)
 
 # Generate a Gaussian latitude axis, north-to-south
 def createGaussianAxis(nlat):
@@ -959,12 +959,40 @@ class AbstractAxis(CdmsObj):
     def isLinear(self):
         raise CDMSError, MethodNotImplemented
 
-    def getBounds(self):
+    def getBounds(self, isGeneric=None):
+        '''
+        isGeneric is a list with one boolean which says if the bounds
+        are read from file (False) or generated (True)
+        '''
         raise CDMSError, MethodNotImplemented
 
-    # Return None if not explicitly defined
     def getExplicitBounds(self):
+        '''
+        Return None if not explicitly defined
+        This is a way to determine if attributes are defined at cell
+        or at point level. If this function returns None attributes are
+        defined at points, otherwise they are defined at cells
+        '''
         raise CDMSError, MethodNotImplemented
+
+    def getBoundsForDualGrid(self, dualGrid):
+        '''
+        dualGrid changes the type of dataset from the current type to the dual.
+        So, if we have a point dataset we switch to a cell dataset and viceversa.
+        '''
+        explicitBounds = self.getExplicitBounds()
+        if (explicitBounds is None):
+            # point data
+            if (dualGrid):
+                return self.getBounds()
+            else:
+                return None
+        else:
+            # cell data
+            if (dualGrid):
+                return None
+            else:
+                return explicitBounds
 
     def setBounds(self, bounds):
         raise CDMSError, MethodNotImplemented
@@ -1279,7 +1307,8 @@ class AbstractAxis(CdmsObj):
         The stride k can be positive or negative. Wraparound is
         supported for longitude dimensions or those with a modulus attribute.
         """
-        fullBounds = self.getBounds()
+        isGeneric = [False]
+        fullBounds = self.getBounds(isGeneric)
         _debug=0
         _debugprefix="SS__XX subaxis "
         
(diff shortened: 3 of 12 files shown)


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/126

danlipsa:
@aashish24 @doutriaux1 Please review. Note the 190 modified baselines :-)

aashish24:
the new baselines look beautiful. Could you also post a image from uvcdat 1.5.1 or earlier for comparison? (just pick one for validation)

danlipsa:
Seems like old images are also noisy.
uvcdat-1 5 1

danlipsa:
Seems like old images are also noisy. This is for uvcdat 1.5.1
uvcdat-1 5 1

doutriaux1:
@danlipsa that actually worries me a bit, why aren't the new ones noisy? As I mentioned on my comments on the baselines PR I'm not sure if it's good or bad I just want to understand why it is so drastically different. The are prettier but I'm more concerned about them being right :wink:

danlipsa:
@doutriaux1 A way to look at this is: before this change = nearest neighbor interpolation - simpler, more noisy. after this change = linear interpolation - looks at all neighbors and averages their value based on the distance to the point, smoother. Given that we reconstruct a continuous signal out of points, neither is right :-) but I think the later has more properties we want and I would argue this is what people expect. Besides, it nicely simplifies the code as well.

jbeezley:
I can provide some insight from the modeling perspective.  Unfortunately, the correct way to do interpolation will be highly dependent on the source data and how it is represented in the model.

The atmospheric models I have worked on in the past represent data for most variables at the cell center (cell values in VTK parlance).  The exception to this is for vector components such as wind velocity and fluxes.  The numbers these represent are normal components to the cell's face positioned at the center.  It is possible that cdms does some regridding to normalize the values that vcs receives, so it would be a good idea to look at what is done there.

In terms of interpolation methods, the most correct method depends on the data type and ideally on how it is interpolated within the model.  Categorical fields should not be interpolated bilinearly for instance.  When regridding categorical fields are usually "accumulated" to the dominant value in the cell, but in this case nearest neighbor would probably be sufficient.  Flux fields should be interpolated using a "mass preserving" interpolation method (the integrated flux should remain the same).  Other fields often need to be interpolated within the model in a way that is globally differentiable.

This is all to say that the modelers will have concerns beyond just what is the most visually pleasing.  In my experience, if you are going to treat all of the data the same, it is best just to use nearest neighbor interpolation to stay true to the actual data in the file.

doutriaux1:
@jbeezley thanks for the input that is coherent with the general feedback I'm getting so far.

danlipsa:
@jbeezley @doutriaux1 Some clarifications:
1. There is no actual nearest neighbor interpolation in the code. This was my way of simplifying things. Cell centers are used as points when we generate the grid for plots that need point data. This way we get a slightly smaller grid than what is in the nc file. We do generate the right grid when we do plots that need cell data. I think we should generate the same grid regardless of what plot we need and this is what this change does.
2. We use linear interpolation either way to generate the isofill/isoline.

Good questions:
1. What kind of interpolation should we use? Nearest neighbor produces bad artifacts for many vis algorithms and it is not differentiable as you have discontinuities when you cross neighbor boundaries.
2. Is the high frequency signal we see in the images before the change features of the data or artefacts we generate because we use a different grid than what is specified in the data.

aashish24:
> 1. There is no actual nearest neighbor interpolation in the code. This was my way of simplifying things. Cell centers are used as points when we generate the grid for plots that need point data. This way we get a slightly smaller grid than what is in the nc file. We do generate the right grid when we do plots that need cell data. I think we should generate the same grid regardless of what plot we need and this is what this change does.

@danlipsa can you elaborate what you meant by "same grid"? 

Thanks,

danlipsa:
@aashish24 You can replace 'same grid' with 'same VTK dataset' to be more precise. After we generate the data we can apply the appropriate conversion for the plot we want. The current implementation, generates directly the converted data - but it uses a simplified conversion, that I think generates the artifacts we see. That also has the disadvantage that it uses two different dataset generation routines which leads to code duplication and opportunities for bugs.

aashish24:
So @danlipsa and I talked and @danlipsa is going to look into the code some more before we have finally say on what grid type we need to generate which is logically correct.

danlipsa:
@doutriaux1 So, with @aashish24 we looked at ncdump output for clt.nc. That does not have bounds information, so you would think this would produce a dataset with information stored on points. Looking at the code, cdms2 generates bounds if they are not in stored in the file - see axis.py::_autobounds. We also generate bounds in vcs2vtk.py::getBoundsList. So, it seems that we think most datasets would have information stored at cell center, even if that is not specified in the file. Do you know why is that?

danlipsa:
@doutriaux1 @aashish24 I added a plot_based_dual_grid=True option, that gives us the current behavior.
We can also turn that option off and plot the grid that is given by the netcdf file. Here are the tests results.
Please review.

https://open.cdash.org/index.php?compare1=63&filtercount=2&field1=buildname%2Fstring&project=UV-CDAT&field2=buildstarttime%2Fdate&showfilters=0&limit=100&compare2=83&value1=0d6bfec7&showfeed=0&value2=20160510T101543

aashish24:
@danlipsa great.  the variable name "plot_based_dual_grid" is bit confusing. What is dual grid means here?

danlipsa:
@aashish24 I am open to suggestions for a name for that variable. :-). For the dual grid definition see the attached image:

aashish24:
thanks for the snapshot @danlipsa why its called dual though? Dual grid means something else to me.

danlipsa:
@aashish24 Not sure it is a common usage. I remember being using in this context when I did the cam5 catalyst adapter. There is a dual graph of a triangulation which is very similar with this.
https://graphics.stanford.edu/courses/cs268-09-winter/notes/handout5.pdf

What does it mean to you? Also feel free to suggest a different word.

danlipsa:
@doutriaux1 @aashish24 Ready to merge?

doutriaux1:
@danlipsa sorry didn't realize you were done with this one. Reviewing today.

danlipsa:
@doutriaux1 @aashish24 ping?

doutriaux1:
oh yes, got distracted by @aashish24 PR but since it's not passing, let me do this one now.

aashish24:
> oh yes, got distracted by @aashish24 PR but since it's not passing, let me do this one now.

@doutriaux1 which PR you are referring to?

doutriaux1:
#1968

doutriaux1:
@danlipsa some failures, re-running to post to dash board:

```
	187 - test_vcs_basic_boxfill_zero (Failed)
	248 - test_vcs_basic_boxfill_aeqd_proj_SH (Failed)
	358 - test_vcs_boxfill_lev1_lev2 (Failed)
	401 - test_vcs_issue_960_labels (Failed)
	419 - vcs_test_fewer_colors_than_levels (Failed)
	430 - test_vcs_no_continents (Failed)
```

danlipsa:
@doutriaux1 uvcdat-testdata needed to be rebased as well. I pushed that. These are my failures, none seems related:
	 54 - test_vcs_read_old_scr_2 (Failed)
	 65 - test_vcs_verify_boxfill_basics (Failed)
	 66 - test_vcs_verify_proj_basics (Failed)
	 67 - test_vcs_verify_fillarea_basics (Failed)
	430 - test_vcs_no_continents (Failed)
	562 - diags_test_02 (Failed)
	563 - diags_test_03 (Failed)
	564 - diags_test_04 (Failed)
	565 - diags_test_41 (Failed)
	566 - diags_test_05 (Failed)
	567 - diags_test_06 (Failed)
	568 - diags_test_07 (Failed)
	569 - diags_test_08 (Failed)
	570 - diags_test_09 (Failed)
	571 - diags_test_10 (Failed)
	572 - diags_test_11 (Failed)
	573 - diags_test_12 (Failed)
	574 - diags_test_13 (Failed)
	575 - diags_test_15 (Failed)

doutriaux1:
https://open.cdash.org/viewTest.php?onlyfailed&buildid=4368763

danlipsa:
@doutriaux1 Did you see my latest message about merging uvcdat-testdata? Can you rerun your tests with the new uvcdat-testdata?

doutriaux1:
yep re-runnig the tests now

doutriaux1:
@danlipsa https://open.cdash.org/viewTest.php?onlyfailed&buildid=4369973 Do you want me to add the missing baseline?

Packages/cdms2/Lib/grid.py
```diff
@@ -496,14 +496,6 @@ def genBounds(self):
...
 
-        # Stretch latitude bounds to +/- 90.0
-        if ascending:
-            latbnds[0,0] = min(latbnds[0,0],-90.0)
-            latbnds[-1,1] = max(latbnds[-1,1],90.0)
-        else:
-            latbnds[0,0] = max(latbnds[0,0],+90.0)
-            latbnds[-1,1] = min(latbnds[-1,1],-90.0)
-
         # Get longitude bounds
```
doutriaux1: @danlipsa why did you remove this?

danlipsa: With the new _genericBounds_flag, I got a different average for a few tests.
The reason was that bounds were generated in grid.genBounds, and modified in the offending code which I removed, instead of axis.genGenericBounds. Also see the following changed baseline (first image).
https://github.com/UV-CDAT/uvcdat-testdata/pull/126/files
Image 4 does not have the purple lines which I think should not be there - all those images are different kind of interpolations.

danlipsa: See also the comment from the commit.

danlipsa:
> @danlipsa https://open.cdash.org/viewTest.php?onlyfailed&buildid=4369973 Do you want me to add the missing baseline?
@doutriaux1 That would be great! Thanks!

doutriaux1:
@danlipsa I pushed the new file. Do you mind running the one test on your Linux box to see if it passes there as well or if we need a _1 file.

danlipsa:
@doutriaux1 Great. Thanks! I'll run the tests and report the results.

danlipsa:
@doutriaux1 test_vcs_no_continents passes on my machine.

PR #1952, opened 2016-04-26
Cdscan importable

doutriaux1:
@mcenerney1 @painter1 this is for you

couldn't quite let you do
`import cdscan`
you will need to do:
`from cdms2 import cdscan`


painter1:
Thank you.  It works now!  I've already updated the devel branch of uvcmetrics, and will issue a pull request to put it in the master branch.

PR #1956, opened 2016-04-28
Fix continents=0

chaosphere2112:
Fixes an issue where plotting without continents would break stuff indefinitely. Test inbound.

### DIFF
--- a/Packages/vcs/vcs/VTKPlots.py
+++ b/Packages/vcs/vcs/VTKPlots.py
@@ -755,7 +755,10 @@ class VTKVCSBackend(object):
                 plot.onClosing(cell)
 
     def plotContinents(self, wc, projection, wrap, vp, priority, **kargs):
-        contData = vcs2vtk.prepContinents(self.canvas._continentspath())
+        continents_path = self.canvas._continentspath()
+        if continents_path is None:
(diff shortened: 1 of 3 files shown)


chaosphere2112:
@doutriaux1 @aashish24 @danlipsa Please review. Bumped into this while doing an example for someone, should go in 2.4.2.

aashish24:
LGTM :+1: you meant 2.6? I don't think we talkd about 2.4.2?

aashish24:
ping @chaosphere2112 just wanted to make sure you saw my +2

chaosphere2112:
@aashish24 I was confused about a UVCMetrics thing; UVCMetrics is getting a 2.4.2, but since it's now decoupled from UV-CDAT's release cycle, we won't be having a corresponding release. Should go in 2.6, though. Sorry, I've been out sick for a few days.

aashish24:
> @aashish24 I was confused about a UVCMetrics thing; UVCMetrics is getting a 2.4.2, but since it's now decoupled from UV-CDAT's release cycle, we won't be having a corresponding release. Should go in 2.6, though. Sorry, I've been out sick for a few days.

No problem. thanks for the update. this branch is good to go if you want to merge it.

Issue #1740, opened 2015-12-17
canvas.png writes a PNG with labels off if bg=0


danlipsa:
Running this test
[test2.txt](https://github.com/UV-CDAT/uvcdat/files/65625/test2.txt)
produces
![bad](https://cloud.githubusercontent.com/assets/6180727/11874925/0eff8468-a4b0-11e5-959e-5c3d8b778f14.png)
Uncommenting the line that sets bg to 1 fixes the problem.

danlipsa:
@aashish24 @doutriaux1 Is this important? Should we fix it before the release?

doutriaux1:
actually yes! Lots of people run things in batch mode. Is it going to be long? Let's see if we can get the other PR in before we move to this one!

doutriaux1:
@jbeezley do you think you can help with this one? Looks like it's `bugging` you as well.

jbeezley:
Probably not, I know almost nothing about VTK.

doutriaux1:
Is there such a person at Kitware? :wink: No worries.

chaosphere2112:
@danlipsa @doutriaux1 This is the issue that I filed ages ago with VTK; we have it as #1148.

aashish24:
@danlipsa I can look into it after I finish my sprint tasks let me know.

aashish24:
thanks @chaosphere2112 for the reference.

danlipsa:
@aashish24 @doutriaux1 @chaosphere2112 I looked at this briefly when I filed is. The reason why the labels are off seems to be that  the display happens on a very large screen which is not available if bg=0. There these events that have to be dealt with properly: ModifiedEvent and ConfigureEvent. This is in VTKPlots.py

That is not to say that there is not a VTK bug somewhere as well :-) I have not seen Sam's bug report at all when I looked at this.

danlipsa:
The case when you don't specify a window size (as the test is written) has been fixed by:
https://github.com/UV-CDAT/uvcdat/pull/1821
When we specify a window size bigger than the screen size:
x=vcs.init()
x.open(4000, 3000)
we still produce the wrong result.

danlipsa:
This might be fixed when we switch VTK to OpenGL2 - a new, much faster, rendering backend in VTK

doutriaux1:
@danlipsa is that coming soon?

danlipsa:
[...]

PR #1967, opened 2016-05-12
Reenable autot_axis tests. Keep max Y to 500.

danlipsa:


danlipsa:
Data in
https://github.com/UV-CDAT/uvcdat-testdata/pull/131

danlipsa:
@doutriaux1 @aashish24 Please review

doutriaux1:
@danlipsa is it ok now on mac? The original comment said it crashes on mac, why is it not crashing anymore?

testing/vcs/test_vcs_autot_axis_titles.py
```diff
@@ -18,8 +18,8 @@
 plot = sys.argv[3]
 x_over_y = sys.argv[4]
 if (x_over_y == '0.5'):
-    xSize = 400
-    ySize = 800
+    xSize = 250
+    ySize = 500
 else:
     xSize = 800
     ySize = 400
```
doutriaux1: don't you need `250` and `500` here as well?

danlipsa:
@doutriaux1 The tests worked on my mac because I had higher resolution. I switched to the lowest resolution and some of the tests failed - I fixed that by reducing the size of the images generated. I think the problem is related to
https://github.com/UV-CDAT/uvcdat/issues/1740

testing/vcs/test_vcs_autot_axis_titles.py
danlipsa: The problem way the Y size was bigger than the screen size. The way it is now works fine on the lowest resolution on my mac: 1024 x 640

danlipsa:
@doutriaux1 @aashish24 ping

testing/vcs/test_vcs_autot_axis_titles.py
aashish24: @danlipsa what is the logic of picking these hard-coded x and y size?

danlipsa: @aashish24 I am making the window tall and narrow or long and thin so that ratio=autot is will be very different than the old behavior where the data is scaled with the window.

aashish24:
LGTM 👍

PR #1983, opened 2016-05-19
BUG: Fix memory override for vtkContourFiler in isofillpipeline.

danlipsa:
This resulted in vtkStripper to generate double coverage of isocountours which resulted in
messed-up patterns.
Also adjusted plot patterns to easier to discriminate.

### DIFF
--- a/Packages/vcs/vcs/vcs2vtk.py
+++ b/Packages/vcs/vcs/vcs2vtk.py
@@ -1581,16 +1581,16 @@ def __build_ld__():
 
 def stippleLine(prop, line_type):
     if line_type == 'long-dash':
-        prop.SetLineStipplePattern(int('1111111100000000', 2))
+        prop.SetLineStipplePattern(int('0000111111111111', 2))
         prop.SetLineStippleRepeatFactor(1)
     elif line_type == 'dot':
-        prop.SetLineStipplePattern(int('1010101010101010', 2))
+        prop.SetLineStipplePattern(int('0101010101010101', 2))
         prop.SetLineStippleRepeatFactor(1)
     elif line_type == 'dash':
-        prop.SetLineStipplePattern(int('1111000011110000', 2))
+        prop.SetLineStipplePattern(int('0001111100011111', 2))
         prop.SetLineStippleRepeatFactor(1)
     elif line_type == 'dash-dot':
-        prop.SetLineStipplePattern(int('0011110000110011', 2))
+        prop.SetLineStipplePattern(int('0101111101011111', 2))
         prop.SetLineStippleRepeatFactor(1)
     elif line_type == 'solid':
         prop.SetLineStipplePattern(int('1111111111111111', 2))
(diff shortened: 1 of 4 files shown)


danlipsa:
Data in:
https://github.com/UV-CDAT/uvcdat-testdata/pull/133

danlipsa:
@doutriaux1 @aashish24 please review.

doutriaux1:
I'm excited about this one!

aashish24:
what was the logic behind:

danlipsa:
This writes over the end of the array.
Not sure what was the original intention.

aashish24:
Yeah it seems weird, I couldn't think of why it was like this. 

LGTM :+1:

aashish24:
@danlipsa you want to use regression module. 

so instead of this

```
+
 +x = vcs.init(bg=1, geometry=(1620, 1080))
 +x.setantialiasing(0)
 +x.drawlogooff()
```

You can do:

x = regression.init(bg=1, geometry=(800, 600)) # Let's use 800x600 from now on 
....
..
..
regression.run(x, "filename")
```

danlipsa:
@aashish24 Should we do 800 x 600 even for background images? That seems very small considering that we can make images as big as we want.

aashish24:
@danlipsa if we make images smaller for testing, then we will have faster checkout of baselines. In vtk the sizes are 400, 400.

doutriaux1:
@danlipsa @aashish24 I don[t think making the picture smaller will make the test much faster, but we will  lose precision in the output and we might not be able to catch small difference, in things like dotted line and such. Just my two cents.

danlipsa:
@aashish24 @doutriaux1 Indeed, for this particular test, 800 x 600 does not catch the difference between a wrong file and a good one - the difference is about 5.

danlipsa:
For the current resolution the difference is 20 is the error is caught.

danlipsa:
@aashish24 Should we merge this in? Note the many sizes we currently have in the repo:
250 x 500
400 x 768
400 x 800
500 x 250
500 x 500
800 x 400
800 x 600
814 x 303
814 x 606
814 x 628
1091 x 1200
1200 x 1090
1200 x 1091
2400 x 2182

aashish24:
@danlipsa in that case can we use 1200 x 1091 (the default size in regression.py) unless you don't think it catches the error. I see your point that we have different size for testing right now but most of them are using regression.py so it would be nice if we can use 1200x1091.

aashish24:
but I am fine if the current resolution is best for the testing  too :+1:

danlipsa:
@doutriaux1 @aashish24 1200 x 1091 is fine. In that case we should look at the baselines of size 1200 x 1090 and make sure we don't compare them against images of size 1200 x 1091.

danlipsa:
@doutriaux1 @aashish24 Note line stipples are not supported yet in OpenGL2 backend for VTK because they have been removed from OpenGL 3.1.
http://www.vtk.org/Bug/view.php?id=15799

Issue #1537, opened 2015-09-10
Enable postscript export to write text objects in addition to text as paths (shapes)


durack1:
As discussed in #1527 the current `postscript` export function writes text objects as shapes, rather than encoded as text objects.

It would be preferable to expose the ability to export text as text/string objects, which will provide better control when importing `postscript` output files into vector graphics editing software (such as Adobe Illustrator).

Further discussions are included in #1527 - in particular [here](https://github.com/UV-CDAT/uvcdat/issues/1527#issuecomment-139002911)

@dlonie can describe what needs to be exposed - a potential upside is that file sizes may be reduced (#517)

dlonie:
Just to comment on a point made in #1527, the "text object" mode should _not_ be the default. Leave "text as paths" as the default. The original VTK vector export implementation in VCS exported text objects and it caused problems.

In particular, some vector formats do not support alignment/rotation of text, and embedding the objects results in portability issues as fonts vary greatly across viewers and platforms. To keep things consistent, we decided to export as paths, since that avoids all of these issues.

As for filesize, every large ps/pdf I've been sent to inspect shows the vast majority of the file to be comprised of high-resolution polygonal data from the plot, while text took up a negligible amount of space. The only way around that is to rasterize the plot into the image or downsample the data prior to exporting.

doutriaux1:
@dlonie I was thinking of simply add it to the psotscript function, so it would be on by default for postscript only. I will test with @durack1 as is and if it creates too many issue will put it back as OFF by default. Does that sound good?

durack1:
@doutriaux1 I think @dlonie was recommending leaving the default as is, but exposing the write text as `text object` functionality using an additional keyword argument.. That way all the work to get `ps` files looking good across multiple systems doesn't go to waste.. And only when you want to dig deeper for `text object` control would a user use it..

dlonie:
What @durack1 said :-)

durack1:
@danlipsa @doutriaux1 @aashish24 this would seem to be a very simple addition to the export postscript function - I'd be happy to kick the tyres on this once a branch is in place to test.

aashish24:
@dlonie if we switch to opengl2 backend, would that be giving us text objects or still shapes?

dlonie:
Either -- See vtkGL2PSExporter::SetTextAsPath(bool). If true, all text will be exported as paths. If false, 2D non-math text will be embedded as a text object (3D or mathtext strings will still be paths, though).

[...]

Issue #1527, opened 2015-09-09
Enable exporting plot directly into PPT


aashish24:
- research if this is possible

durack1:
@aashish24 just a question.. What renderer does `VTK` use for `PS` exports? I was trying to edit fonts in a VCS-exported `*.ps` file and realised that the fonts didn't appear to be created correctly - they aren't text objects in the (very large) exported file..

aashish24:
there is a module for GL2PS in vtk. @dlonie would know more about that. @dlonie can you point @durack1 to the source code?

durack1:
@aashish24 thanks, will await hints from @dlonie 

For what it's worth, as a UV-CDAT user I would never use an option to directly export to a `*.pptx` file, as there is very little control exposed in that software (color space etc)..

Here are some relevant links:
http://graphicdesign.stackexchange.com/questions/22423/why-does-illustrator-cs6-save-for-microsoft-office-use-png-format
https://en.wikipedia.org/wiki/Windows_Metafile (Hybrid Windows vector/bitmap format - looks like `v12.0` was just released in 2015, so could be another output format to support)
https://en.wikipedia.org/wiki/Microsoft_Open_Specification_Promise#Graphics_formats

dlonie:
Text in VCS vector graphics is exported as a set of paths (that is, the characters are drawn as shapes in the file, rather than encoded as strings). This is to work around a number of export and portability issues regarding alignment, viewer font availability, portable regression testing, etc. If you'd like to modify them, you'll need to erase the path objects and replace them with a new text object.

durack1:
@dlonie ok, I was under the impression that text objects would be more like:
```
/Times-Roman findfont
12 scalefont
setfont
newpath
100 200 moveto
(Example 3) show
```
Rather than the alternate.. Is there a way of labeling the shapes as text objects, so that an import to vector editing software (e.g. Adobe Illustrator) will know about it being text? It would certainly aid the ability to tweak figures after exporting..

dlonie:
We had it like that, but changed it to use paths as not all of the vector formats support things like text alignment/rotation, and fonts vary greatly from platform to platform. These issues leading to portability problems and bad-looking outputs.

It should be possible to add an option to the export methods in VCS to toggle between text objects and path exports. Check with @doutriaux1 about it, as it would require some API changes in the VCS library.

durack1:
[...]

PR #1992, opened 2016-05-26
Added option to export text as object or path

aashish24:


aashish24:
@durack1 these are actually ps files but since github does not allow ps to be uploaded, I changed them to pdf extension. Could you please look into these two? Make sure to change the extension to ps after you download them. 

[foo_as_object.pdf](https://github.com/UV-CDAT/uvcdat/files/284521/foo_as_object.pdf)
[foo_as_paths.pdf](https://github.com/UV-CDAT/uvcdat/files/284522/foo_as_paths.pdf)

aashish24:
Ref https://github.com/UV-CDAT/uvcdat/issues/1537

aashish24:
@durack1 it would be great if you can look at this soon for 2.6. I will add a test and if you are happy with the API and results, I will merge it.

durack1:
@aashish24 thanks.. I personally like having the text objects, which then import very neatly into graphics packages.. I also note that the file sizes are markedly different - with the ***text object file 5KB***, whereas the ***text paths file is 103KB*** this would be a major issue in much more complex figures.

The text objects also display more cleanly on my mac, at least for your example files:
160526_vectortextoutput

I do wonder whether some of this needs tweaking though, it seems that you're plotting the text objects/paths outside of a bounding box (so compare the left side of the fig above to the file opened using a vector graphics package below):
foo_as_object_il

durack1:
@aashish24 I also wonder whether you've managed to grab all the formats - so `pdf`,`ps`,`eps` and `svg` should all have an option to export text as object - [you might be missing the `eps`](https://github.com/UV-CDAT/uvcdat/blob/0b2065ad031cf1e0b9c4efc1006273a65647d1f6/Packages/vcs/vcs/Canvas.py#L5330-L5331)

Also what is the changed syntax for calling this new functionality?

aashish24:
> I do wonder whether some of this needs tweaking though, it seems that you're plotting the text objects/paths outside of a bounding box (so compare the left side of the fig above to the file opened using a vector graphics package below):

@durack1 that's just my sample script and the code that decides whats in / out is not controlled by my changes.

aashish24:
> pdf,ps,eps and svg

pdf, ps, and svg is all covered. I didn't see eps (is that ghostscript?).

aashish24:
> @aashish24 thanks.. I personally like having the text objects, which then import very neatly into graphics packages

Cool! Yes text as object is the default like you wanted :smile:

aashish24:
@durack1 I have to add a test but other than that if you like everything else then I am going to go ahead with adding some test for it.

durack1:
@dlonie can you remind @aashish24 and myself why the export text as paths was the default? And while you're at it, can you suggest any fringe test cases that should be included to make sure that the proposed change doesn't trip things over

dlonie:
This is mainly a rehash of what I said in #1527 and #1537, but maybe it'll clear things up?

The output formats supported by gl2ps (which VTK uses for postscript/pdf/svg/etc vector exports) handle text objects inconsistently. For example, postscript mangles newlines, pdf doesn't fully support rotation and alignment, stuff like that. These are limitations in the actual format specifications themselves.

On top of that, embedding text objects then relies on the viewer to locate a similar font and render the text, and odds are good that the fonts used by the viewer will have different characteristics than the ones used in the original rendering. So, for instance, you have some right-justified lines of text, like the data at the top of the VCS plots. If the font used by the viewer uses different widths for *any* of glyphs composing the text, the text will be unaligned along the right-hand side, since the text is always anchored on it's left side due to how these formats represent text objects. This just looks bad.

Exporting text as paths eliminates all of these problems with portability across viewers and inconsistent text object handling between output formats.

For testing, it's somewhat pointless to do in this case, IMO. The differences between backends (e.g. no newlines in postscript) will always be there when text is exported as objects -- text objects cannot correctly represent the rendering in that case. Same with aligned rotations in pdf. As for portability differences, you can't fully test for that, as different platforms/viewers will use different fonts, and they'll always look bad for certain cases (e.g. right-justified text). Trying to regression test using system-dependent fonts is a sure-fire way to have unstable and unreliable tests.

My recommendation is the same as in #1527 -- leave paths as the default, and allow objects as an option. This way things will work for the 99% of users who just want an accurate reproduction of their rendering, and the 1% that wants to tweak the text after exporting will still be able to easily.

doutriaux1:
I agree  with @dlonie , Also @chaosphere2112 can we copy paste this excellent explanation on our wiki. @aashish24 this explanation should also go in the vector format output doc strings.

durack1:
@doutriaux1 ditto from me for every statement you made above. I think the description from @dlonie should be everywhere.. And his defaults suggestion too..

durack1:
@aashish24 @doutriaux1 I also think a documentation of file size implications is also useful info for documenting.. My 5KB vs 103KB is a starter very basic example above

durack1:
@dlonie just a quick thanks for taking the time to write that description above.. It's certainly very useful to all of us to have your input!

Packages/vcs/vcs/Canvas.py
```diff
@@ -5312,19 +5313,19 @@ def gif(self, filename='noname.gif', merge='r', orientation=None,
...
 
     ##########################################################################
-    #                                                                           #
-    # Screen Encapsulated PostScript wrapper for VCS.                           #
-    #                                                                           #
+    #                                                                        #
+    # Screen Encapsulated PostScript wrapper for VCS.                        #
+    #                                                                        #
     ##########################################################################
     def eps(self, file, mode='r', orientation=None, width=None, height=None, units='inches',
```
durack1: @aashish24 this also needs a `, textAsObject=True` appended

dlonie:
Happy to help.

aashish24:
This discussion is very useful. I made the path as default with the option ``textAsPaths`` set to ``True``. User can override this option when they call the method. Also I documented the reasoning behind this option in vectorGraphics method. I felt that's the best place since the reasoning seems to be implementation dependent. I will add explanation for it my rst documentation as well.

aashish24:
One another thing, I found the eps method was broken, I fixed it too. @durack1 it would be nice if you can verify the results for all formats.

durack1:
@doutriaux1 is there a way to get a branch installed using conda - so theoretically we have nightly, but how do I best test branches that are not in `master`?

doutriaux1:
@aashish24 while it's nice to put the explanation in vectorgraphics function, we need to add it in the doc string of pdf/postscript/eps/svg, etc... These are the functions that the end user will access.

aashish24:
I will update the rst_docs and the method post 2.6

durack1:
@aashish24 thanks for this.. As a user I would really like inline docs to include documentation on this change, so I would advocate for a doc update **prior** to tagging 2.6 rather than after

aashish24:
I totally understand @durack1. The document I am referring to is much detailed documentation which is not only for this option for other options as well. I just pushed and merged a basic vcs user-guide but I am afraid I won't have deep dive user-guide for 2.6. What I can do in the mean-time is document in "changes in 2.6" for the end-users.

danlipsa:
@aashish24 See the following flake8 warnings. Is this you? :smile: 
[~/projects/uvcdat/build]$ /home/danlipsa/projects/uvcdat/build/install/bin/flake8 "/home/danlipsa/projects/uvcdat/src/Packages/vcs/vcs/" "--show-source" "--statistics" "--max-line-length=120"
/home/danlipsa/projects/uvcdat/src/Packages/vcs/vcs/dv3d.py:162:5: E303 too many blank lines (2)
    def setProvenanceHandler(self, provenanceHandler):
    ^
/home/danlipsa/projects/uvcdat/src/Packages/vcs/vcs/Canvas.py:4917:5: E303 too many blank lines (2)
    def postscript(self, file, mode='r', orientation=None, width=None, height=None,
    ^
2       E303 too many blank lines (2)

aashish24:
> def setProvenanceHandler(self, provenanceHandler):

mostly likely not me



> def postscript(self, file, mode='r', orientation=None, width=None, height=None,

Might be me. Do you want me to push a fix?

danlipsa:
@aashish24 That would be great. Thanks.

doutriaux1:
@aashish24 @danlipsa I fixed this in my conda_build branch

durack1:
@aashish24 just so we're all on the same page there are VCS docs available from http://uvcdat.llnl.gov/documentation/vcs/vcs.html - it'd be great to refresh and update these docs as they are the most comprehensive currently available.

I also have a preference for inline docs with functions - so using the dir(function) or print function.__doc__ approach.. This would make it really easy for a command line VCS user to be exposed to all the new functionality without having to figure out where to point a web browser

danlipsa:
@doutriaux1 The problem is that we have a failing flake8 until your conda branch is merged. As a process, I would rather have someone who breaks master fix it, rather than many people fixing that in their branches, or waiting for a particular unrelated branch until master is fixed. What do you think?

danlipsa:
@aashish24 or anyone else could push a fix in 5 min, instead of waiting a day for your conda branch or any other branch.

doutriaux1:
@danlipsa I don't really care, but I really care about the conda branch being merged in sooner rather than later.

PR #1998, opened 2016-05-31
Revert commit that break dv3d_ tests

danlipsa:
The commit is 9404c34157ededc830a4287d35e6082e55edf6bf

### DIFF
--- a/Packages/vcs/vcs/utils.py
+++ b/Packages/vcs/vcs/utils.py
@@ -1629,7 +1629,7 @@ def getgraphicsmethod(type, name):
     return copy_mthd
 
 
-def creategraphicsmethod(gtype, gname='default', name=None):
+def creategraphicsmethod(gtype, name):
     if gtype in ['isoline', 'Gi']:
         func = vcs.createisoline
     elif gtype in ['isofill', 'Gfi']:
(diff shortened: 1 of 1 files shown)


danlipsa:
@doutriaux1 @aashish24 Reverting the listed commit fixes the dv3d_ tests. Please review.

doutriaux1:
@danlipsa the new GUI actually needs this. @sampsonbryce @chaosphere2112 do you mind fixing the places where dv3d calls this? Or we can merge this and you guys can fix it later.

chaosphere2112:
@danlipsa @doutriaux1 I'll push a fix to dv3d, I know the root of the problem.

danlipsa:
@chaosphere2112 Thanks! I will close this then.

aashish24:
@chaosphere2112 could you please elaborate why this change was required? Also, I am wondering how did we merge this one? Did we not tests?

chaosphere2112:
@aashish24 The "required" change was fixing the way VCS addons were handled (I use a histogram in the new GUI). My intern took the DV3D fix into his own hands to fix what looked like a hole in the API from when it was integrated (and indeed, was). @doutriaux1 got a little merge-happy (and thought I had run tests, which I had not), so that's how it got merged.

There were actually an existing host of problems related to this; any time you plotted a 3D graphics method, it used the *exact same* graphics method object, even when plotted on multiple canvases. This meant that modifications to the first canvas (like panning the camera) would effect the second canvas. While this could be useful in some circumstances, it totally breaks the existing beliefs about how UV-CDAT works. I'm fixing that problem now, as part of fixing these tests (which were badly written, and depended on examining things in `canvas.backend`, which is a no-fly zone for API compatibility as far as I'm concerned).

ThomasMaxwell:
dv3d was designed such that navigation commands (pan, zoom, rotate, etc) would be synchronized across all cells.   Having to adjust the camera independently in each cell is extremely awkward when doing 3D visualization.

PR #2000, opened 2016-05-31
Fix regression for test_vcs_png_window_resize

danlipsa:
The regression is from a7f5b860ff1408282431b70890044f956396c320


danlipsa:
@aashish24 This went through from you vectors branch.

PR #2005, opened 2016-06-02
BUG: Extend the isoline attribute list with the last value from the e…

danlipsa:
…xisting attribute list

If there are no attributes in the list, we choose
default values: linewidth=1, linestyle='solid', linecolor='black'

### DIFF
--- a/Packages/vcs/vcs/vcsvtk/isolinepipeline.py
+++ b/Packages/vcs/vcs/vcsvtk/isolinepipeline.py
@@ -14,6 +14,14 @@ class IsolinePipeline(Pipeline2D):
         super(IsolinePipeline, self).__init__(gm, context_)
         self._needsCellData = False
 
+    def extendAttribute(self, attributes, default):
+        if len(attributes) < len(self._contourLevels):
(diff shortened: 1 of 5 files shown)


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/139

Packages/vcs/vcs/vcsvtk/isolinepipeline.py
```diff
@@ -47,11 +53,19 @@ def _plotInternal(self):
 
         if len(linewidth) < len(self._contourLevels):
             # fill up the line width values
-            linewidth += [1.0] * (len(self._contourLevels) - len(linewidth))
+            if (len(linewidth) == 0):
```
aashish24: @danlipsa looks to me that this code is repetitive. Would it be useful to have a method in the same class?

danlipsa:
@aashish24 Great suggestion. See the revised code.

aashish24:
LGTM :+1: thanks for considering the suggestion.

PR #2007, opened 2016-06-03
BUG: datawc does not work on a time axis.

danlipsa:


danlipsa:
https://github.com/UV-CDAT/uvcdat-testdata/pull/140

danlipsa:
@doutriaux1 @aashish24 Please review.

aashish24:
@danlipsa can we not write comparison operator for time axis? 

http://stackoverflow.com/questions/5824382/enabling-comparison-for-classes

doutriaux1:
cdtime and cdtime object have a comparison operator (cmp)

danlipsa:
@aashish24 Can you overload the comparison operator to compare to float as well as to cdtime.reltime? Note also that cdtime.reltime is C code.

aashish24:
@danlipsa you cannot have multiple definitions of operators since python is type agnostic but what you can do is something like here: 

```
class B(object):
    def __init__(self, bar):
        self.bar = bar
    def __eq__(self, other):
        if isinstance(other, B):
            return self.bar == other.bar
        elif isinstance(other, A):
            return self.bar + 3 == other.foo
        else:
            return NotImplemented
    def __ne__(self, other):
        result = self.__eq__(other)
        if result is NotImplemented:
            return result
        return not result
```

I would define this operator in axis class since that is in python. Referenced from here: 
http://jcalderone.livejournal.com/32837.html

The advantage might be that once defined in axis.py, the comparison will work everywhere and hence reduced code duplication.

danlipsa:
@aashish24 Thanks for the pattern to override operators. Still, the problem we have to solve is to compare gm.datawc_ (either a float or the type returned by 'cdtime.reltime') with 1.e20. So it seems we would have to override = for cdtime.reltime which is C code.

Not sure I would like to hide through an operator the fact that datawc can have different types. I would rather investigate why datawc needs that, and make it only be float. @doutriaux1 Do you know what datawc gets converted to cdtime.reltime?

doutriaux1:
@danlipsa it gets converted to relative time to accommodate for plotting data with different time units.

For example 2016-06-06 is 157 days since 2016-01-01 but is 5 days since 2016-06-01

So if you plot two datasets with the time but different untis, we need to convert to some "same units" (and calendar) time format ot be able to align the times properly. Does that make sense?

danlipsa:
@doutriaux1 Thanks for the explanation.

danlipsa:
> So if you plot two datasets with the time but different untis, we need to convert to some "same units" (and calendar) time format ot be able to align the times properly.

@doutriaux1 Do we have a test for this?

doutriaux1:
you could refactor these:
```
doutriaux1@omar:[/git/uvcdat/Packages/vcs/Test]:[pattern_scaling_aspect]:[7957]> ll *datawc*
-rw-rw-r-- 1 doutriaux1 doutriaux1  904 Mar 21 08:09 test_boxfill_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1  932 Mar 21 08:09 test_isofill_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1  896 Mar 21 08:09 test_isoline_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1 1.2K Mar 21 08:09 test_scatter_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1  905 Mar 21 08:09 test_vector_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1 1.2K Mar 21 08:09 test_xvsy_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1  947 Mar 21 08:09 test_xyvsy_datawc_time.py
-rw-rw-r-- 1 doutriaux1 doutriaux1  933 Mar 21 08:09 test_yxvsx_datawc_time.py
```

danlipsa:
@doutriaux1 @aashish24 is OK with merging this as is. If you are OK as well, I'll merge it.

Packages/vcs/vcs/utils.py
```diff
@@ -1666,13 +1666,25 @@ def creategraphicsmethod(gtype, gname='default', name=None):
     return copy_mthd
 
 
+# Returns the float value for datawc_...
+# datawc_ can be a float or a cdtime.reltime
+# TODO: Investigate why datawc is converted to a cdtime.reltime
+def getDataWcValue(v):
+    if (type(v) is type(cdtime.reltime(0, 'months since 1900'))):
```
doutriaux1: @danlipsa why `months since 1900`, we should use the `gm.datawc_units` and `gm.datawc_calendar`

danlipsa: @doutriaux1 This is a way to get the type, so the actual value does not matter. I got this from Canvas.py:3051. Is there a better way to get the type returned by cdtime.reltime?

doutriaux1: what about component time?
"""python
a=cdtime.reltime(45,"months since 1945")
b=cdtime.comptime(2000)
gm = x.createboxfill()
gm.datawc_x1 = b
type(gm.datawc_x1) is type(a)
```

danlipsa: @doutriaux1 What is comptime? I did not try it, so probably not. It does not work with rel time either - I did try that. Its not clear from the documentation that datawc_ accepts objects beyond float values. I can added an issue for this and we can fix it and add tests.
https://github.com/UV-CDAT/uvcdat/issues/2009
The question is: should we do this for 2.6 or leave it for later? Note this PR solves the user's issue. @aashish24 ?

doutriaux1: ok. A `comptime` is a componentTime. Basically a date. Whereas `reltime` is a relative time, i.e a delta since some reference date. You should take a look at: http://uvcdat.llnl.gov/documentation/cdms/cdms_3.html

aashish24:
@doutriaux1 @danlipsa few things: 

1. What the zone of componentTime? GMT?
2.  One thing I found confusing is that in this case, user has entered float value and then we convert it into reltime (on the same variable). I think it would be better
2 (a): For conversion to store in a separate variable/ 
2 (b): Not modify the original value and type as it leads to bugs / confusion for new users.

doutriaux1:
@aashish24 there's no local in comptime, it's a good suggestion, we might want to add this.

2. is a good suggestion, actually we should do it for all vcs elements, there's a lot of conversion under the hood that leads to oiginal user values being changed. example "dot" becomes a number, etc... Let's open a PR for this issue.

danlipsa:
@doutriaux1 @aashish24 Should we merge this? We can clean-up and fix additional related datawc/time axis bug in subsequent PRs.

Issue #1947, opened 2016-04-25
isofill vcs bug


doutriaux1:
using the attached file Dave was able to produce the following plot. The max is 29.xx so no data should be in the pink 30-35 interval.

![isofillbugmaybe](https://cloud.githubusercontent.com/assets/2781425/14788142/7cb63ca6-0abb-11e6-844a-59edcad3085c.png)

[HadSST1870to99.nc.zip](https://github.com/UV-CDAT/uvcdat/files/234803/HadSST1870to99.nc.zip)

doutriaux1:
hum... I'm not sure what's going on here. I was able to reproduce exactly Dave's plot with:
```python
import cdms2
import vcs

x=vcs.init()
f=cdms2.open("HadSST1870to99.nc")
s=f("sst")
print s.min(),s.max()
iso=x.createisofill()
iso.levels=range(-5,36,5)
x.plot(s,iso)
raw_input("Press enter")
```

but using:
```python
import cdms2
import vcs

x=vcs.init()
f=cdms2.open("HadSST1870to99.nc")
s=f("sst")
print s.min(),s.max()
iso=x.createisofill()
iso.levels=range(-5,34,5)
x.plot(s,iso)
raw_input("Press enter")
```

I get:
![good](https://cloud.githubusercontent.com/assets/2781425/15911854/68e39172-2d86-11e6-8905-1f354df5de13.png)

doutriaux1:
@danlipsa @aashish24 I think I'm near the issue it seems that we don't actually care what values the user pass see this code:
```python
            cots.append(cot)
            mapper.SetInputConnection(cot.GetOutputPort())
            lut.SetNumberOfTableValues(len(tmpColors[i]))
            for j, color in enumerate(tmpColors[i]):
                r, g, b, a = self.getColorIndexOrRGBA(_colorMap, color)
                if style == 'solid':
                    tmpOpacity = tmpOpacities[j]
                    if tmpOpacity is None:
                        tmpOpacity = a / 100.
                    else:
                        tmpOpacity = tmpOpacities[j] / 100.
                    print "DO WE COME HERE",j,r,g,b,tmpOpacity
                    lut.SetTableValue(j, r / 100., g / 100., b / 100., tmpOpacity)
                else:
                    print "OR DO WE COME HERE"
                    lut.SetTableValue(j, 1., 1., 1., 0.)
            print "WE COME HERE"
            luts.append([lut, [0, len(l) - 1, True]])
[...]

PR #2027, opened 2016-06-12
BUG #1947: isofill does not handle out of bounds levels correctly.

danlipsa:
When smallest level is bigger than min scalar value or biggest level
is smaller than max scalar value isofill creates the wrong image.


danlipsa:
@doutriaux1 @aashish24 Please review.

danlipsa:
@doutriaux1 The confusing part here was that we color by the cell data generated by vtkBandedPolyDataContourFilter. These are the bands from 0 to len(l) - 1. But we can also color by the scalar in the original data - this is point data: this is what you've done in  your PR. Your PR looked correct, but I did not understand initially how this was working before. I forgot about the bands generated by vtkBandedPolyDataContourFilter.

doutriaux1:
@danlipsa did you also test the last example I posted, I don't think this fixes it. 
```python
import cdms2
import vcs

x=vcs.init()
f=cdms2.open("HadSST1870to99.nc")
s=f("sst")
print s.min(),s.max()
iso=x.createisofill()
iso.levels=[-15,-14,-13,-12,-11,-10,-9,-8,-7,-6,35]
x.plot(s,iso)
```

danlipsa:
@doutriaux1 I run a selection in ParaView for all cells >= -6. This selects most cells - note ParaView does not show blanked cells because I save polygonal data which does not support blanking.

isofill-interval


This is the image generated by uvcdat - if you look carefully you'll see different colors at the top.
isofill-interval

danlipsa:
@doutriaux1 Actually, I think you are right, there is something going on here. Some black should have been white. See the following image:
isofill-interval

doutriaux1:
@danlipsa yes that is odd. But at least the levels seem to be picked up correctly (on your first plot). I still don't see how it's possible from the code, but it seems to work.

doutriaux1:
@danlipsa I merged `conda_build` into your branch (locally) and was able to test it quickly, it seems to work ok. Not sure about the black bit though as you mentioned.

danlipsa:
@doutriaux1 @aashish24 Please review. This solves all problems I know of.

doutriaux1:
@danlipsa looks like it's working now. Thanks. But why do we do a `pointtocell` ? 
The reason I'm asking is that I was hoping we could draw isofill/isolines on unstructured grid, even if they did not have vertices information, simply using a point dataset.

danlipsa:
@doutriaux1 putMaskOnVTKGrid needs to draw black (masked) any cell that has any points hidden. That is why I need pointtocell.

You won't be able to do isolines/isofill for a point dataset. Both need interpolation which only works if you have cells - topology information. Otherwise the algorithm won't know which points to use for interpolation.

doutriaux1:
@danlipsa I'm attaching a sample zipped nc file, that has coordinate info (lat/lon) but no bounds info (in CF format), vcs cannot plot it, but it should be able to, since we know the point data (lat/lon). It's definitely something for 3.0 anyway, just wanted to ask you about it now before I forget.
[add.zip](https://github.com/UV-CDAT/uvcdat/files/324157/add.zip)

doutriaux1:
@danlipsa I updated the baselines, it's good to be merged in as far as I'm concerned. If it's good with you as well, go ahead and merge please.

PR #2032, opened 2016-06-20
needed to duplicate yr fully in case the year starts in October for e…

doutriaux1:
…xample

@aashish24 @danlipsa @chaosphere2112 it's an easy one.


aashish24:
LGTM :+1:

PR #2044, opened 2016-07-01
Change default for setbgoutputdimensions units to pixels from inches.

danlipsa:


danlipsa:
@doutriaux1 @aashish24 Easy one to review.

aashish24:
thanks @danlipsa can you also change the test suite (remove passing "pixels" arg to the relevant method). ?

danlipsa:
@aashish24 I made the suggested changes.

danlipsa:
@aashish24 @doutriaux1 ping

danlipsa:
Do: test

doutriaux1:
@danlipsa I was looking at this yesterday with @chaosphere2112  on a RH6 machine, and this is totally ignored. The only way to get the png dimension to change these days is to use the `width` argument in the `x.png()` call. I think we should get rid of this function all together (or ignore it and put a deprecation warning). Also x.setbgoutputdimension() and x.plot(bg=0) was seg faulting n a RH6 yesterday...

danlipsa:
@doutriaux1 @aashish24 in that case, should we get rid of bgX and bgY as well? I think we hadsome problems related to these v. So we should always use geometry['width'] and geometry['height']. Maybe we need to add another variable to Canvas that tells us if we do background rendering or not.

aashish24:
+1 for removing ``setbgoutputdimensions``. I think @danlipsa if you remember we talked about it too as it will clean up the code and confusing API as well.

doutriaux1:
@danlipsa yes I think it's all useless and confusing nowdays.

aashish24:
Great.. it sounds like we all agree that we can get rid of bgdimensions functions call and member variable. @danlipsa should we close this PR then and create a new one with cleaned up API?

danlipsa:
@aashish24 Sure, I'll close this and submit another one.

PR #2065, opened 2016-07-27
Proposed new color names

aashish24:


aashish24:
@doutriaux1 @danlipsa I would like to propose refactoring the color names. The current color names unfortunately were inconsistent and hard to read. I followed the color names syntax as I found in some other tools specifically in R domain (https://www.nceas.ucsb.edu/~frazier/RSpatialGuides/colorPaletteCheatsheet.pdf) but I am open for suggestions. This in preparation for 3.0 release so, yes it will break the backward compatibility, but if you all insist, I can add a alias for now as well. 

Also, over all the names got shorter, I deleted 277 characters and added 257. 

Thoughts?

doutriaux1:
@aashish24 yes, let's add an alias. Also let's add matplotlib palettes especially `viridis` because I want to switch to viridis for our default in a separate PR.

doutriaux1:
@aashish24 I will add viridis and other matplotlib colormaps

aashish24:
> @aashish24 I will add viridis and other matplotlib colormaps

@doutriaux1 if its okay, let me add that (and alias) so that we can have one branch for color maps.

aashish24:
thanks @doutriaux1 so sounds like these changes are liked which is great. I think it will really help boost our usability as well are making the API better

doutriaux1:
i was going to do it in that branch but go ahead. Thanks!

aashish24:
> i was going to do it in that branch but go ahead. Thanks!

roger that

aashish24:
Do: test

sankhesh:
Do: test

sankhesh:
Do: check

aashish24:
Do: test

aashish24:
Do: test

aashish24:
@doutriaux1 @danlipsa this should be ready for review

aashish24:
Do: test

aashish24:
Also, added matplotlib colors (plasma, inferno, magma, viridis)

aashish24:
Do: test

Packages/vcs/vcs/Canvas.py
```diff
@@ -4961,6 +4979,15 @@ def setcolormap(self, name):
         #   updateVCSsegments_flag = args[1]
         # except:
         #   updateVCSsegments_flag = 1
+        # Python < 3 DeprecationWarning ignored by default
```
doutriaux1: @aashish24 I don't think this should be here. We should issue this warning whenever the user sets the colormap instead.

doutriaux1:
@aashish24 also the initial.attributes in Share is just ridiculously long, we should reformat it to a straight json dump.

doutriaux1:
@aashish24 let's also use this PR to switch the default colormap to `viridis` do you want me to do this?

Packages/vcs/vcs/Canvas.py
aashish24: thanks. @doutriaux1 I am bit confused. Currently the warning is in "setcolormap" instead. Look at the code at line 4953.

aashish24:
> @aashish24 also the initial.attributes in Share is just ridiculously long, we should reformat it to a straight json dump.

we could do that. I made a pass so that its pretty. Even though its long, its easy to read. I am fine either way.

aashish24:
> @aashish24 let's also use this PR to switch the default colormap to viridis do you want me to do this?

@doutriaux1 okay if we do it in a separate PR to keep pr's small?

Packages/vcs/vcs/Canvas.py
doutriaux1: graphic methods can have colormaps as well, is that going to be triggered if I do:
```python
import vcs
b = vcs.createboxfill()
b.colormap = "white_to_red"
```

doutriaux1: remember there are 3 levels to look for the colormap used.
1- graphics method
2- canvas
3 vcs module

doutriaux1:
that's fine by me

aashish24:
@doutriaux1 I uglified the initial attributes as suggested, that made the file size three times smaller, I liked your idea, thanks. This branch is ready for review now.

aashish24:
Do: test

aashish24:
Do: test

aashish24:
Do: test

aashish24:
@doutriaux1 @danlipsa @williams13  this is ready for review.

Packages/vcs/vcs/Canvas.py
aashish24: > remember there are 3 levels to look for the colormap used.
> 1- graphics method
> 2- canvas
> 3 vcs module

Okay, let me add something to the code to make sure that we can map old to new regardless of where we set the colormap.

danlipsa:
@aashish24 The new names LGTM 👍

sankhesh:
Do: test

aashish24:
@doutriaux1 @danlipsa okay so I updated the code so that it will be backward compatible no matter how we set the colormap. Please have a look at the code now. Its ready for the review :black_joker:

aashish24:
@doutriaux1 @danlipsa ping

danlipsa:
@aashish24 LGTM 👍

aashish24:
Do: test

aashish24:
> @aashish24 LGTM  i would squash the commits though

@danlipsa each commit is different and brings new stuff to the code base. I am not entirely sure that in this particular case squashing makes sense. Actually if you look the branch there are only 4 commits as I rebased so the older commits are gone if that's what you were referring to.

danlipsa:
@aashish24 Why would you have 'Added backward compatibility support' and 'Updated backward compatibility' as commits by themselves? Also, why would you have 'Proposed new color names' and 'Updated to use new color names' as separate commits?

doutriaux1:
@aashish24 does `graphic_method.colormap = "old_name" issues the warning?

aashish24:
> @aashish24 Why would you have 'Added backward compatibility support' and 'Updated backward compatibility' as commits by themselves? Also, why would you have 'Proposed new color names' and 'Updated to use new color names' as separate commits?

the first commit added the backward compatibility in a separate way which is different than the last commit on the same topic. I kept old to keep that history of what things I have tried. The proposed color names just added new names and the updated to use new color names updated tests to use new names vs the old ones. I am fine with squashing this into one.

aashish24:
> @aashish24 does graphic_method.colormap = "old_name" issues the warning?

@doutriaux1 no it does not. I thought about it and I don't think with direct assignment like this we can issue a warning. Now we can issue a warning afterwards when we actually try to use the color name but it may not be as useful (especially since we user colormap at different places and very differently). Managing that could be lot of work. I think we document it and send an email to the mailing list that may suffice to inform the user but I am open for ideas.

doutriaux1:
@aashish24 it should be a `property` so you can issue the warning in the `setter` function

aashish24:
> @aashish24 it should be a property so you can issue the warning in the setter function

I see that but then I guess I have to create a override function for all graphics method (there is no common baseclass?) Its more core (canvas and all other graphics methods etc..) but i am fine with that if that is more useful.

aashish24:
thanks @doutriaux1. I am wondering if we should have a base class for graphics methods. Thoughts? @doutriaux1 @danlipsa

doutriaux1:
I've been thinking about i for a while, there's something similar to that in vcsaddons, we inspire ourselves from it.

PR #2077, opened 2016-08-08
Fixes #2061

### LINKED ISSUE #2061: vcs.displayplot.Dp.script() does not generate .json ouput
In the source for the script() function for the Dp class, it looks like the function can only ever output a Python object.

embrown:
@chaosphere2112 Apparently I decided to send that nuke via snailmail rather than ICBM. Still, it got there eventually.


embrown:
This fix was included in a PR that was already merged.

PR #2086, opened 2016-08-11
Fix bug with orthographic projection

chaosphere2112:
This makes it so the `centerlongitude` attribute on orthographic projections actually works. Not entirely clear why the existing code didn't behave appropriately, `"lon_0"` is right according to the proj4 documentation, but this works right in VCS. Goes with UV-CDAT/uvcdat-testdata#151


danlipsa:
LGTM :+1: Seems the issue was that CentralMeridian overwrote lon_0 that we were setting.

doutriaux1:
@chaosphere2112 test_center_longitude is missing frm repo

doutriaux1:
@chaosphere2112 wait for new baselines to be updated and then submit PR with corrected picture please.

chaosphere2112:
@doutriaux1 Yeah, it's missing from my hard drive as well. I'll rewrite the test and submit.

PR #2087, opened 2016-08-12
Fix for template name collision bug.

jccosta:
Fix for avoiding name cached for templates created by EzTempla and TemplateFactory.

### DIFF
--- a/Packages/vcsaddons/Lib/Multi.py
+++ b/Packages/vcsaddons/Lib/Multi.py
@@ -229,7 +229,7 @@ class Multi(object):
             self.x.removeP(t)
         self.template_names=[]
         
-    def get(self,column=None,row=None,legend=None,font=True,fontlimit=0.8):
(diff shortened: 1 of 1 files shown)


jccosta:
Removed whitespaces.

jccosta:
Updated alignment and removed all whitespace (emacs says so. (-: )
This is a way to increase the "uniqueness" of the template names and so avoid cached names to collide.

Packages/vcsaddons/Lib/Multi.py
```diff
@@ -229,7 +229,7 @@ def clean(self):
             self.x.removeP(t)
         self.template_names=[]
         
-    def get(self,column=None,row=None,legend=None,font=True,fontlimit=0.8):
+    def get(self,column=None,row=None,legend=None,font=True,fontlimit=0.8,tname=None):
```
doutriaux1: `tname` is not a great name but I'm fine with it.

PR #2096, opened 2016-08-22
Vcs color utility

aashish24:


aashish24:
@doutriaux1 @danlipsa I added this small web-based utility to visualize colors as defined in initial.attributes. i am open for suggestions.  Few things in my mind: 

1. Can add python-interface to visualize matplotlib colors as well. 
2. Can find where the initial.attributes file is. 
3. Can find color data from vcs (instead of initial.attributes). 

I don't want to spend too much time but anything that require minimal time can be done.

doutriaux1:
@aashish24 why js? We should have done it in python/vcs, no?

aashish24:
@doutriaux1 JS because in the future, I want to make it interactive as well (use can create new VCS color maps for example). I just though that web version would be so much nicer for interactive purposes. Also, we can make it super pretty in the future if we use web. 

Python would have been nice too but I can slap a python interface on top of it in the future as well.

doutriaux1:
Then I feel you should submit this to the vcdat repo.

aashish24:
@doutriaux1 this is to help the vcs documentation and has nothing to do with vcdat or vcs. It's a very small utility code that will help us with the color map documentation in vcs.

doutriaux1:
I'll let @embrown and @chaosphere2112 decide then, @embrown can we have sphinx use this? Otherwise I think it should be in Python and in vcs/utils.py

aashish24:
okay, in that case I move this to a separate repo. Any objection in creating a new repo under UV-CDAT umbrella?

chaosphere2112:
@aashish24 Can you just put it in UV-CDAT/vcs.js?

aashish24:
> @aashish24 Can you just put it in UV-CDAT/vcs.js?

we could do that too. I am fine with that. Let me close this PR and I will another one for vcs.js.

PR #2115, opened 2016-08-31
Export all graphics (including patterns) as vectors

### LINKED ISSUE #2072: export patterns / hatches as vectors for vector formats

sankhesh:
Fixes #2072

### DIFF
--- a/Packages/vcs/vcs/VTKPlots.py
+++ b/Packages/vcs/vcs/VTKPlots.py
@@ -70,7 +70,6 @@ class VTKVCSBackend(object):
         # Turn on anti-aliasing by default
         # Initially set to 16x Multi-Sampled Anti-Aliasing
         self.antialiasing = 8
-        self._rasterPropsInVectorFormats = False
         self._geometry = geometry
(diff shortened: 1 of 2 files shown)


sankhesh:
Do: test

aashish24:
@sankhesh any progress with autotrace? I looked into the inkscape and the results were very good. Let's talk about it today.

sankhesh:
@aashish24 Sure. The only obstacle is hinted and anti-aliased text now. 
That should be done in VTK with libfreetype.

aashish24:
Just wanted ot update everyone that we are looking for an alternate solution that will minimize the file size but it will require some work to create the patterns as geometry in the first place. More update on this soon.

aashish24:
@sankhesh I am going to close this PR as we are likely to follow the geometry path all the way.

doutriaux1:
@aashish24 actually I think it's the way to go @chaosphere2112 was actually suggesting that we should pass fillarea object as patterns. That would be awesome.

PR #2124, opened 2016-09-26
Additional test for BUG #5: vtkPolyData::RemoveDeletedCells does not …

danlipsa:
…remove GlobalIds or points.


danlipsa:
@doutriaux1 @aashish24 This is an additional test for the vcs isofill fix. Please review.

danlipsa:
Do: test

danlipsa:
@doutriaux1 I should add the previous tests from datawc commit as well isn't it?

doutriaux1:
I think so

doutriaux1:
@danlipsa almost all good, but in diags the vectors disappeared, see:
https://open.cdash.org/testDetails.php?test=487168860&build=4573138

danlipsa:
@doutriaux1 I'll take a look.

danlipsa:
@doutriaux1 I updated to the latest branch but I get:
[~/projects/vcs/src (fix_isofill_levels %=)]$ python setup.py install --old-and-unmanageable
  File "setup.py", line 10, in <module>
    import cdat_info
ImportError: No module named cdat_info

doutriaux1:
@danlipsa any progress on this? I would like to merge it in soon

danlipsa:
@doutriaux1 I should be able to push a fix soon. The problem is similar with what I fixed before but slightly trickier - I cannot set the hidden vectors to a value because they'll still show up. I try to just delete the points.

danlipsa:
@doutriaux1 Feel free to apply your commits to the master branch. This way you won't depend on my PR. Your commits and my PR should be independent.

danlipsa:
@doutriaux1 @aashish24 An update on this. Here is a picture that shows the hidden points in red. I remove the cells around the hidden points, but note that there are some points left in the hidden area.
hidden-points
This causes some additional arrows. Those did not show up in the past, I think because of the default hidden point behavior which I think deletes the cells around a hidden point and also deletes all cell points.
Here are the two pictures: test result:
figure-set6_global_ann_stress_plot-combined
baseline:
figure-set6_global_ann_stress_plot-combined
I can restore the old behavior but I wanted to point this out.

doutriaux1:
@danlipsa I think this is fine, but can you push this so that we see the new baselines differences on cdash? It's easier with the slider.

danlipsa:
@doutriaux1 @aashish24 I pushed the fix. All tests pass in dignostics - probably the difference I pointed out is to small. A new baseline was changed in uvcdat.

danlipsa:
For your reference, these are the additional PRs:
UV-CDAT/uvcdat-testdata#155
https://github.com/UV-CDAT/vcs/pull/10

PR #2144, opened 2016-12-01
Added test for script() with .py extension

embrown:
Test for https://github.com/UV-CDAT/vcs/pull/104
@doutriaux1 Did I do this correctly?

### DIFF
--- a/testing/vcs/test_script_for_py.py
+++ b/testing/vcs/test_script_for_py.py
@@ -1,0 +1,15 @@
+import vcs
+
+list=["boxfill","meshfill","isofill","textcombined","texttable","vector"]
+for obj in list:
+    s=''
+    tc=''
+    if obj == 'textcombined':
+        vcs.createtextcombined('EXAMPLE_tt', 'qa', 'EXAMPLE_tto', '7left')
+        tc="'EXAMPLE_tt','EXAMPLE_tto'"
+    code="s=vcs.get%s(%s)" % (obj, tc)
+    exec(code)
+    try:
+        s.script(obj+"_script.py")
+    except:
+        raise("Python script creation broke on "+obj+".script()")
--- a/testing/vcs/CMakeLists.txt
+++ b/testing/vcs/CMakeLists.txt
@@ -1197,6 +1197,12 @@ cdat_add_test(test_vcs_textextents
   ${BASELINE_DIR}/test_textextents.png
 )
 
+cdat_add_test(test_script_for_py
+  "${PYTHON_EXECUTABLE}"
+  ${cdat_SOURCE_DIR}/testing/vcs/test_script_for_py.py
+)
+
+
 
 
 add_subdirectory(vtk_ui)


aashish24:
Looks correct to me but I didnot run the code

testing/vcs/test_script_for_py.py
```diff
@@ -0,0 +1,15 @@
+import vcs
+
+list=["boxfill","meshfill","isofill","textcombined","texttable","vector"]
```
chaosphere2112: You shouldn't name things reserved words in python (`list` being the type of `[]`). It technically works but is bad form, and will break things later down the road.

testing/vcs/test_script_for_py.py
```diff
@@ -0,0 +1,15 @@
...
+for obj in list:
+    s=''
+    tc=''
+    if obj == 'textcombined':
+        vcs.createtextcombined('EXAMPLE_tt', 'qa', 'EXAMPLE_tto', '7left')
+        tc="'EXAMPLE_tt','EXAMPLE_tto'"
+    code="s=vcs.get%s(%s)" % (obj, tc)
+    exec(code)
```
chaosphere2112: It's bad form to use exec for... well, anything. I know it's sprinkled throughout the codebase, but you shouldn't pick up bad habits from that :wink: 

Instead, you can use this:

```
s = vcs.creategraphicsmethod(obj, 'test_%s' % obj)
```

which will let you generate an arbitrary graphics method.

testing/vcs/test_script_for_py.py
chaosphere2112: Also, you're missing a few:

```
"isoline", "1d", "taylordiagram", "3d_scalar", "3d_dual_scalar", "3d_vector"
``` 

Remove "textcombined" and "texttable"; these should be handled in a second loop, which will do secondary objects (marker, fillarea, colormap, line, textcombined, texttable, textorientation).

testing/vcs/test_script_for_py.py
```diff
@@ -0,0 +1,15 @@
+import vcs
+
+list=["boxfill","meshfill","isofill","textcombined","texttable","vector"]
+for obj in list:
+    s=''
+    tc=''
+    if obj == 'textcombined':
```
chaosphere2112: Let's extract the secondary graphics methods to a separate loop. Also drop the below logic.

testing/vcs/test_script_for_py.py
```diff
@@ -0,0 +1,15 @@
...
+        vcs.createtextcombined('EXAMPLE_tt', 'qa', 'EXAMPLE_tto', '7left')
+        tc="'EXAMPLE_tt','EXAMPLE_tto'"
+    code="s=vcs.get%s(%s)" % (obj, tc)
+    exec(code)
+    try:
+        s.script(obj+"_script.py")
```
chaosphere2112: You need to delete the files after you generate them. Also, this only builds the script; for this to be a real test, it needs to validate correctness. That means running the scripts and seeing if the generated object matches the original.

testing/vcs/test_script_for_py.py
```diff
@@ -0,0 +1,15 @@
...
+    except:
+        raise("Python script creation broke on "+obj+".script()")
```
chaosphere2112: This doesn't work, for a couple of reasons:

- raise is a keyword, not a function. This is just a misleadingly worded version of `raise "Python script creation broke on..."`.
- You can only raise objects that derive from the `BaseException` class in python. This usually means something like: `raise ValueError("Python script creation broke on...").

Moreover, you should just `print "Python script creation broke on..."` and do `sys.exit(1)`.

embrown:
@chaosphere2112, thank you for the review! I'll get working on the revisions.

embrown:
@chaosphere2112 @doutriaux1, got the test written, but it looks like I don't have push access to the uvcdat repo any more. If anyone's here Tuesday I'll get admin help and push it then.

doutriaux1:
@embrown i think iot's because travis ci fails then olny admins can approve.

doutriaux1:
pretty uch every test fails there... Let me take a look...

PR #2151, opened 2017-01-11
Adds doctesting to the VCS test suite.

embrown:
Currently only queries test is implemented.
Once https://github.com/UV-CDAT/vcs/pull/114 is merged, this test
will pass. Tests for other modules/classes will be very similar.

Already did some setup for running with vcs.Canvas. A few of the
steps in this script aren't necessary for queries, but will be
needed for others (the cleanup step is mainly for Canvas and
manageElements).

If docstring contains '.. pragma: skip-doctest', the doctest
for that docstring will not be run.
If docstring is empty and not private, and has not been skipped,
it will be printed in a "NO DOCUMENTATION" list after the tests
so we can view it on cdash, but it will not cause a failure.

### DIFF
--- a/testing/vcs/test_vcs_docstrings_queries.py
+++ b/testing/vcs/test_vcs_docstrings_queries.py
@@ -1,0 +1,65 @@
+from vcs import queries
+import vcs.utils
+import glob
+import os
+import doctest
+
+
+def cleanup():
+    """Cleanup for the doctests. If some files aren't being deleted after testing, add their glob signature to the
+    patterns list.
+    """
+    gb = glob.glob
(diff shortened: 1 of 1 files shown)


embrown:
@doutriaux1 @chaosphere2112 I got the bulk of the modules/classes in VCS. 

Tests that break:
* meshfill
* isofill
* textcombined
* texttable

Every test will break until https://github.com/UV-CDAT/vcs/pull/114 is merged. The above tests, and some tests I have not added yet, will break until https://github.com/UV-CDAT/uvcdat/pull/2144 and https://github.com/UV-CDAT/vcs/pull/104 are merged.

PR #2179, opened 2017-05-15
fix #2178

### LINKED ISSUE #2178: ask anonymous should periodically ask again in case it's no

doutriaux1:


doutriaux1:
@danlipsa @aashish24 feel free to review
@williams13 I set it up to ask again every month if set to not sending data, is that too frequent?
It also ask again in case you upgraded to a greater version.

PR #2205, opened 2018-04-03
fixed v8.0 yaml files

muryanto1:
updated v8.0 yaml files


doutriaux1:
@muryanto1 how did you generate the new yml. I'm a bit nervous about changing so many files. In general if a yaml is broken we should go to great length to keep the same version of everything in place, updating any package version (like new python here) might introduce new bugs or features that were not present with older versions.

